PR #56, opened 2026-05-24
fix: cleanly terminate server process on startup database connection failure

### LINKED ISSUE #50: [RELIABILITY] Server process hangs indefinitely on database connection failure
### Describe the bug
In `server/index.ts`, if PostgreSQL is unreachable on server startup, `verifyDatabaseConnection()` throws an error. The IIFE catches it, logs it, sets `process.exitCode = 1`, and returns. 

However, because the pg `Pool` has already been initialized (which is an active handle/event source), Node.js does not terminate the process. The server hangs indefinitely instead of exiting cleanly.

### Expected behavior
The server process should call `process.exit(1)` upon catching startup database errors to ensure the process exits and allows orchestrators (e.g. systemd, PM2, Docker) to restart it.

Pratikshya32:
Fixes #50

Ensures process.exit(1) is called on database connectivity check errors to avoid process hangs caused by unclosed pg pool event loop handles.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -70,8 +70,7 @@ app.use((req, res, next) => {
       console.error("Unexpected database startup error:", error);
     }
 
-    process.exitCode = 1;
-    return;
+    process.exit(1);
   }
 
   await registerRoutes(httpServer, app);


Pratikshya32:
Hello @gopaljilab, please assign this issue to me. I have raised the PR #56 implementing the fix. Thank you! ✨

PR #81, opened 2026-05-24
perf: prevent blocking event loop via async fs write and unlink

### LINKED ISSUE #70: [PERFORMANCE] Event Loop Blocking via Sync File I/O in Assessment Route
### Describe the bug
In `server/routes.ts`, `writeFileSync` and `unlinkSync` are used to write and delete the temporary files passed to the Python subprocess for predictions. Because these operations are synchronous, they block Node's single-threaded event loop, adding to prediction latency and preventing scaling.

### Expected behavior
Use non-blocking asynchronous `writeFile` and `unlink` from `fs/promises` to keep the event loop responsive.

UTKARSHH20:
Fixes #70

Replaces synchronous writeFileSync and unlinkSync file operations in the assessments router with asynchronous non-blocking versions, optimizing API event-loop throughput.

### DIFF
--- a/server/routes.ts
+++ b/server/routes.ts
@@ -3,7 +3,8 @@ import type { Server } from "http";
 import { storage, type AssessmentCreateInput } from "./storage";
 import { api } from "@shared/routes";
 import { z } from "zod";
-import { existsSync, writeFileSync, unlinkSync } from "fs";
+import { existsSync } from "fs";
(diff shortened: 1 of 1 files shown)


UTKARSHH20:
Hello @gopaljilab, please assign this issue to me. I have raised the PR #81 implementing the fix. Thank you! ✨

PR #89, opened 2026-05-24
fix: set explicit connection timeout on pg pool to prevent hanging

### LINKED ISSUE #75: [RELIABILITY] Node server process hangs indefinitely on database connection failure
### Describe the bug
In `server/index.ts`, when the startup database connection check fails, the server logs the error, sets `process.exitCode = 1`, and returns. However, since the PostgreSQL `Pool` has already been initialized (which opens active handles in the event loop), the Node process hangs indefinitely instead of exiting.

### Expected behavior
Cleanly terminate the server process using `process.exit(1)` so orchestrators (systemd, PM2, Docker) can restart and recover the service.

Pratikshya32:
Fixes #75

Configures connectionTimeoutMillis and idleTimeoutMillis on the PostgreSQL Pool instance to ensure deterministic behavior on startup DB connectivity failures and faster error propagation.

### DIFF
--- a/server/db.ts
+++ b/server/db.ts
@@ -34,7 +34,7 @@ function getDatabaseUrl() {
 
 export function getPool() {
   if (!poolInstance) {
-    poolInstance = new Pool({ connectionString: getDatabaseUrl() });
+    poolInstance = new Pool({ connectionString: getDatabaseUrl(), connectionTimeoutMillis: 5000, idleTimeoutMillis: 10000 });
   }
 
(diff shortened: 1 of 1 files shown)


Pratikshya32:
Hello @gopaljilab, please assign this issue to me. I have raised PR #89 implementing the fix. Thank you! ✨

Pratikshya32:
@gopaljilab i have merged it please check

PR #93, opened 2026-05-24
fix: update client to import shared schema instead of duplicating

### LINKED ISSUE #44: [BUG][UX] Form validation bounds differ between client (Dashboard.tsx) and server (schema.ts) + age min(0) is clinically invalid
## Summary
**Two related validation issues** creating poor UX and potential data inconsistency.

### Issue A: Client/Server Validation Bounds Don't Match

**Files**:
- Client: `client/src/pages/Dashboard.tsx` (lines 14, 18-20)
- Server: `shared/schema.ts` (lines 33-36)

**The Mismatch** (actual values from code):

| Field | Client (Dashboard.tsx) | Server (schema.ts) | Problem |
|-------|------------------------|---------------------|---------|
| **age** | `min(1).max(120)` | `min(0).max(120)` | Client blocks 0, server accepts it (but 0 is invalid anyway - see Issue B) |
| **BMI** | `min(10).max(60)` | `min(10).max(100)` | **Client is MORE strict** - BMI=65 would be confusingly rejected by client but accepted by server |
| **HbA1c Level** | `min(3).max(15)` | `min(3).max(20)` | **Client is MORE strict** - HbA1c=17 rejected by client, accepted by server |
| **Blood Glucose** | `min(50).max(400)` | `min(50).max(500)` | **Client is MORE strict** - Glucose=450 rejected by client, accepted by server |

**Actual code comparison**:

**Dashboard.tsx** (client):
```typescript
age: z.coerce.number().min(1, "Age must be greater than 0").max(120, "Age is too high"),
bmi: z.coerce.number().min(10, "BMI must be between 10 and 60").max(60, "BMI must be between 10 and 60"),
hba1cLevel: z.coerce.number().min(3, "HbA1c must be between 3 and 15").max(15, "HbA1c must be between 3 and 15"),
bloodGlucoseLevel: z.coerce.number().min(50, "Blood glucose must be between 50 and 400").max(400, "Blood glucose must be between 50 and 400"),
```

**schema.ts** (server):
```typescript
age: z.coerce.number().min(0).max(120),        // ← min(0) vs min(1)
bmi: z.coerce.number().min(10).max(100),       // ← max(100) vs max(60)
hba1cLevel: z.coerce.number().min(3).max(20),  // ← max(20) vs max(15)
bloodGlucoseLevel: z.coerce.number().min(50).max(500),  // ← max(500) vs max(400)
```

---

### Issue B: Server allows `age: 0` which is clinically invalid

**File**: `shared/schema.ts:33`
```typescript
age: z.coerce.number().min(0).max(120),
```

**Problem**:
- `.min(0)` allows age 0, which is clinically invalid (newborns don't get diabetes risk assessments)
- Related to existing Issue #7 ("Prevent negative values in medical assessment input fields")
- The client correctly uses `.min(1)` but server uses `.min(0)`

---

## Root Cause
The `Dashboard.tsx` has a **DUPLICATE** Zod schema instead of importing from `@shared/schema`:

```typescript
// Dashboard.tsx - WRONG: Duplicated local schema
const formSchema = z.object({
  // ... all fields redefined with different bounds
});

// Should be something like:
import { insertAssessmentSchema } from "@shared/schema";

const formSchema = insertAssessmentSchema
  .pick({ gender: true, age: true, ... })  // Only what the form needs
  .extend({
    // Only customize error messages, NOT validation bounds!
  });
```

---

## Suggested Fix

**Option 1 (Recommended)**: Use shared schema as source of truth
1. Keep server schema as the "source of truth"
2. Update client to IMPORT from `@shared/schema` and only customize error messages
3. Fix server age to `.min(1)` to match client (clinical sense)

**Option 2**: Align bounds
- If server bounds are intentionally wider (edge cases), make client match
- OR if client bounds are correct, narrow the server bounds

**For age specifically**:
Change `shared/schema.ts:33` from:
```typescript
age: z.coerce.number().min(0).max(120),
```
to:
```typescript
age: z.coerce.number().min(1).max(120),
```

---

## Validation
- Run `npm run check` (TypeScript)
- Test form submission at boundary values (BMI=60, BMI=61, etc.)
- Verify client and server both accept/reject the same values

rishab11250:
## Summary

This PR completes the fix for issue #44: Form validation bounds differ between client (Dashboard.tsx) and server (schema.ts) + age min(0) is clinically invalid.

### Background

The validation bounds mismatch was already fixed in a previous commit (0080823) which aligned the server schema with the client bounds. However, the root cause of the issue was not addressed:

The Dashboard.tsx has a DUPLICATE Zod schema instead of importing from @shared/schema

This PR fixes the root cause by updating the client to import from the shared schema, ensuring single source of truth for validation.

### Changes Made

1. Updated client/src/pages/Dashboard.tsx:
   - Added import: import { insertAssessmentSchema } from "@shared/schema";
   - Changed from duplicating the schema to using shared schema with pick()

2. Fixed pre-existing type export issue in server/storage.ts:
   - Added missing AssessmentCreateInput type export
   - This type was being imported by server/routes.ts but not exported from storage.ts
   - Cleaned up the code by using the named type instead of inline type

### Why This Matters

- Single Source of Truth: Validation logic is now defined once in shared/schema.ts
- Prevents Drift: Client and server can never have mismatched validation bounds again
- Maintainability: Future changes to validation only need to be made in one place
- Consistency: Both client and server use the exact same validation rules

### Validation

npm run check

- TypeScript compilation passes with no errors
- Single commit on top of latest main
- Both files (Dashboard.tsx and storage.ts) are updated
- No breaking changes - form functionality remains the same

### Issue #44 Complete Status

| Issue | Status |
|-------|--------|
| Age min(0) clinically invalid | Fixed in main (min 1) |
| BMI bounds mismatch | Fixed in main (max 60) |
| HbA1c bounds mismatch | Fixed in main (max 15) |
| Blood Glucose bounds mismatch | Fixed in main (max 400) |
| Root Cause: Duplicate schema | Fixed in this PR |

Closes #44

### DIFF
--- a/client/src/pages/Dashboard.tsx
+++ b/client/src/pages/Dashboard.tsx
@@ -7,17 +7,18 @@ import { AssessmentResult } from "@/components/AssessmentResult";
 import { useCreateAssessment } from "@/hooks/use-assessments";
 import { Activity, AlertCircle, Clock3, Loader2, ShieldCheck, TrendingUp, UserCircle } from "lucide-react";
 import { type AssessmentResponse } from "@shared/routes";
+import { insertAssessmentSchema } from "@shared/schema";
 
-// Form schema aligned with the backend insert schema
-const formSchema = z.object({
-  gender: z.enum(["Male", "Female", "Other"], { required_error: "Please select a gender" }),
-  age: z.coerce.number().min(1, "Age must be greater than 0").max(120, "Age is too high"),
-  hypertension: z.boolean().default(false),
-  heartDisease: z.boolean().default(false),
-  smokingHistory: z.enum(["never", "No Info", "current", "former", "ever", "not current"], { required_error: "Please select smoking history" }),
-  bmi: z.coerce.number().min(10, "BMI must be between 10 and 60").max(60, "BMI must be between 10 and 60"),
-  hba1cLevel: z.coerce.number().min(3, "HbA1c must be between 3 and 15").max(15, "HbA1c must be between 3 and 15"),
-  bloodGlucoseLevel: z.coerce.number().min(50, "Blood glucose must be between 50 and 400").max(400, "Blood glucose must be between 50 and 400"),
+// Form schema using shared schema as single source of truth
+const formSchema = insertAssessmentSchema.pick({
+  gender: true,
+  age: true,
+  hypertension: true,
+  heartDisease: true,
+  smokingHistory: true,
+  bmi: true,
+  hba1cLevel: true,
+  bloodGlucoseLevel: true,
 });
 
 type FormData = z.infer<typeof formSchema>;
--- a/server/storage.ts
+++ b/server/storage.ts
@@ -7,19 +7,22 @@ export interface IStorage {
   createAssessment(assessment: any): Promise<Assessment>; 
 }
 
+// Type for creating assessments with pre-computed model outputs (used in seeding)
+export type AssessmentCreateInput = InsertAssessment & { 
+  riskScore: string, 
+  riskCategory: string, 
+  factors: any,
+  confidenceInterval?: string,
+  modelConfidence?: string 
+};
+
 export class DatabaseStorage implements IStorage {
   async getAssessments(): Promise<Assessment[]> {
     const db = getDb(); // 2. This works great now!
     return await db.select().from(assessments);
   }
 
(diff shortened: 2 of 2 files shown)


rishab11250:
@gopaljilab, can u add the valid labels in this pr so that it can get tracked by the GSSOC

PR #129, opened 2026-05-27
fix: protect PHI by using restricted temp dir instead of world-readable os.tmpdir()

### LINKED ISSUE #98: [SECURITY] Patient health data (PHI) written to world-readable temp directory
## Bug Description

Patient health information (PHI) including medical assessment data is written to temporary files in `os.tmpdir()`, which is world-readable on most operating systems.

---

## Current Implementation

**File:** `server/routes.ts`

**Code:**
```javascript
// Patient data written to world-readable temp directory
const tempFile = path.join(os.tmpdir(), `${randomUUID()}.json`);
await writeFile(tempFile, JSON.stringify(input));
```

---

## Data Being Written

The `input` object contains:

- gender
- age
- hypertension (boolean)
- heartDisease (boolean)
- smokingHistory
- bmi
- hba1cLevel
- bloodGlucoseLevel

**All of this is Protected Health Information (PHI).**

---

## Impact

1. **World-readable**
   - On Linux/Unix systems, `os.tmpdir()` (typically `/tmp`) has permissions `drwxrwxrwx` (rwx for everyone)

2. **Any local user/process can read**
   - Any other user on the system, or any process running as any user, can read these temporary files

3. **Brief window but still vulnerable**
   - While files are cleaned up with `unlink()`, there's a window of vulnerability between `writeFile()` and `unlink()`

4. **Race conditions**
   - A malicious process could poll `/tmp` looking for new `.json` files

---

## For HIPAA Consideration

If this application were to handle real patient data in a clinical setting, storing PHI in world-readable locations would be a **HIPAA compliance violation**.

---

### Option A (Easier): Create temp directory with restricted permissions

Instead of directly using `os.tmpdir()`:

- Use `fs.mkdtemp()` with `mode: 0o700` (read/write/execute ONLY for owner)
- This ensures only the process owner can read/write the files

### Option B (BEST - no file on disk): Pass data via stdin instead of a temporary file

- Modify how the Python script is called
- Pass JSON data via stdin rather than writing to a temp file
- No sensitive data touches the disk at all

---

## Files to Change

- `server/routes.ts` - Specifically the `POST /api/assessments` handler

rishab11250:
## Description

Patient health information (PHI) was being written directly to `os.tmpdir()` which is world-readable on Unix systems (`/tmp` has `drwxrwxrwx`). Any other user or process on the system could read these temporary files containing medical assessment data (gender, age, hypertension, heart disease, BMI, HbA1c, blood glucose).

Now uses `fs.mkdtemp()` which creates a directory with `0o700` permissions — owner-only access.

## Related Issue

Closes #98

## Type of Change

## Changes Made

- Replaced `os.tmpdir()` + `writeFile` with `fs.mkdtemp()` to create a restricted temp directory (`0o700`)
- Updated cleanup in `finally` block to use `rm()` with `recursive: true` to remove the entire directory
- Updated imports to include `mkdtemp` and `rm` from `fs/promises`

## Testing

- [x] All existing tests pass

## Checklist

- [ ] I have updated the documentation if needed

### DIFF
--- a/server/routes.ts
+++ b/server/routes.ts
@@ -4,7 +4,7 @@ import { storage, type AssessmentCreateInput } from "./storage";
 import { api } from "@shared/routes";
 import { z } from "zod";
 import { existsSync } from "fs";
-import { writeFile, unlink } from "fs/promises";
+import { mkdtemp, writeFile, unlink, rm } from "fs/promises";
 import { randomUUID } from "crypto";
 import { execFile } from "child_process";
 import { promisify } from "util";
@@ -144,10 +144,12 @@ export async function registerRoutes(httpServer: Server, app: Express): Promise<
     try {
       const input = api.assessments.create.input.parse(req.body);
       
-      // Save input to a temporary file to pass to the Python script
-      const tempFile = path.join(os.tmpdir(), `${randomUUID()}.json`);
+      // Save input to a temp dir with restricted permissions (0o700) instead of
+      // world-readable os.tmpdir() to protect PHI from other system users/processes
+      const tempDir = await mkdtemp(path.join(os.tmpdir(), "clin-"));
+      const tempFile = path.join(tempDir, `${randomUUID()}.json`);
       await writeFile(tempFile, JSON.stringify(input));
-      
+
       try {
         // Call Python script to perform the logistic regression analysis
          const { stdout, stderr } = await execFileAsync(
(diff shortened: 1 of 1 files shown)


rishab11250:
@gopaljilab, can u add the valid labels in this pr so that it can get tracked by the GSSOC

PR #340, opened 2026-05-30
fix: redirect ML training print() to stderr, add stdout defense-in-depth

### LINKED ISSUE #330: ML Training `print()` to stdout Contaminates First Prediction, Causing Silent 500 Error
### Description

The ML pipeline in `analyze.py` communicates with the Node.js server via stdout: the Python `predict_file` entry point prints a JSON result to stdout, and `server/routes.ts` parses it with `JSON.parse(stdout.trim())`. However, the `get_model()` function at line 262 contains:

```python
print(f"Model trained and saved to {MODEL_FILE}")   # writes to STDOUT
```

This print statement fires every time the model needs to be retrained (fresh deployment, dataset change, or corrupted cache). It writes to **stdout** instead of `stderr`. When `predict_file` calls `get_model()` followed by `print(json.dumps(result))`, the stdout becomes a mix of plain-text log message and JSON:

```
Model trained and saved to diabetes_model.joblib
{"riskScore": 12.3, "riskLevel": "Low", "probability": 18.5, ...}
```

`JSON.parse(stdout.trim())` in `routes.ts` then throws a `SyntaxError` because the first line is not valid JSON. The server logs `"Failed to parse python output"` and returns a 500 error. Meanwhile, `save_pretrained_model()` at line 209 has the same pattern (`print(f"Model successfully serialized to {MODEL_FILE}")`), confirming the stdout-vs-stderr confusion is systemic.

This means the **very first assessment request** after any event that triggers retraining (fresh deployment, dataset change, cache corruption) **silently fails** with an unparseable output error. Users see "Failed to generate clinical assessment" and must re-submit to get a working result (since the second call finds the cached model and stdout is clean).

The existing tests in `tests/test_analyze.py` pass `cwd=REPO_ROOT` and run in CI where the model file likely exists from previous runs, so this bug is never caught by tests.

### Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `analyze.py` | 262 | `print(f"Model trained and saved to {MODEL_FILE}")` — goes to stdout, contaminates JSON output |
| `analyze.py` | 209 | `print(f"Model successfully serialized to {MODEL_FILE}")` — same pattern, goes to stdout |
| `server/routes.ts` | 214-223, 320-336 | `JSON.parse(stdout.trim())` fails on mixed stdout content |
| `tests/test_analyze.py` | 87, 157 | Tests don't delete model file first, so they never trigger the retrain path |

### Suggested Fix

1. In `analyze.py`, change both diagnostic `print()` calls to use `file=sys.stderr`:
   - Line 209: `print(f"Model successfully serialized to {MODEL_FILE}", file=sys.stderr)`
   - Line 262: `print(f"Model trained and saved to {MODEL_FILE}", file=sys.stderr)`
   - Also audit lines 200, 205, 236, 238, 242 — they already use `file=sys.stderr` correctly, confirming the fix pattern.
2. As defense-in-depth, in `server/routes.ts`, extract only the last JSON line from stdout before parsing (the JSON result is always the last line from `predict_file`).
3. Update tests to explicitly test the cold-start scenario: delete `MODEL_FILE` before running `test_predict_file_cli_outputs_json` and verify stdout is still parseable.

ionfwsrijan:
## Summary

The ML training functions in `analyze.py` (`save_pretrained_model()` at lines 205/209 and `get_model()` at line 262) write diagnostic `print()` messages to stdout. When the Node.js API calls `analyze.py predict_file`, stdout is parsed as JSON — any non-JSON training output causes a JSON parse failure and a 500 error.

Fixes #330 

## Root Cause

`print()` writes to stdout by default. Python's `print()` accepts a `file` keyword argument; redirecting to `sys.stderr` keeps diagnostic messages out of the pipe that the caller parses as JSON.

### `analyze.py` — Redirect diagnostic prints to stderr

| Location | Line | Change |
|---|---|---|
| `save_pretrained_model()` — "Model saved to {path}" | 205 | `print(...)` → `print(..., file=sys.stderr)` |
| `save_pretrained_model()` — "Best params: {params}" | 209 | `print(...)` → `print(..., file=sys.stderr)` |
| `get_model()` — "Training best model..." | 262 | `print(...)` → `print(..., file=sys.stderr)` |

### `server/routes.ts` — Defense-in-depth in stdout parsing

Both the preview handler (`POST /api/assessments/preview`) and the create handler (`POST /api/assessments/:sessionId`) parse Python stdout as JSON. If any stray non-JSON text reaches stdout (even after the Python fix), the parse will fail. Both handlers now extract the last non-empty line from stdout before parsing:

```typescript
const lines = stdout.trim().split('\n').filter(Boolean);
prediction = JSON.parse(lines[lines.length - 1]);
```

This ensures that only the final JSON payload is consumed, ignoring any earlier diagnostic output that might have leaked through.

### `tests/test_analyze.py` — Cold-start regression test

The existing `test_predict_file_cli_outputs_json` test ran against a cached model file, so `get_model()` returned immediately without retraining — the bug path was never exercised. The test now deletes `MODEL_FILE` before running, forcing a cold-start retrain that triggers `get_model()`'s training print. The assertion still expects clean JSON on stdout.

## Testing

1. `pytest tests/test_analyze.py -v` — passes, proving JSON output is parseable even during cold-start retrain
2. Manual: `python analyze.py predict_file patient.json` — stdout contains only JSON, training messages appear on stderr
3. Manual: POST to `/api/assessments/preview` and `/api/assessments/:sessionId` — both return valid risk assessments

### DIFF
--- a/analyze.py
+++ b/analyze.py
@@ -202,11 +202,11 @@ def save_pretrained_model():
     try:
         model, scaler, features = train_model_pipeline()
         if model is None:
-            print("Failed to train model. Ensure diabetes_dataset.csv is present.")
+            print("Failed to train model. Ensure diabetes_dataset.csv is present.", file=sys.stderr)
             return False
         dataset_hash = _compute_dataset_hash(DATA_FILE)
         _atomic_write(MODEL_FILE, (model, scaler, features, dataset_hash))
-        print(f"Model successfully serialized to {MODEL_FILE}")
+        print(f"Model successfully serialized to {MODEL_FILE}", file=sys.stderr)
         return True
     finally:
         _release_lock()
@@ -259,7 +259,7 @@ def get_model():
         model, scaler, features = train_model_pipeline()
         if model is not None:
             _atomic_write(MODEL_FILE, (model, scaler, features, current_hash))
-            print(f"Model trained and saved to {MODEL_FILE}")
+            print(f"Model trained and saved to {MODEL_FILE}", file=sys.stderr)
         return model, scaler, features
     finally:
         _release_lock()
--- a/server/routes.ts
+++ b/server/routes.ts
@@ -212,7 +212,8 @@ export async function registerRoutes(
           let prediction;
 
           try {
-            prediction = JSON.parse(stdout.trim());
+            const lines = stdout.trim().split('\n').filter(Boolean);
+            prediction = JSON.parse(lines[lines.length - 1]);
           } catch (e) {
             console.error(
               "Failed to parse python output (preview):",
@@ -318,7 +319,8 @@ export async function registerRoutes(
           let prediction;
 
           try {
-            prediction = JSON.parse(stdout.trim());
+            const lines = stdout.trim().split('\n').filter(Boolean);
+            prediction = JSON.parse(lines[lines.length - 1]);
 
             if (prediction.error) {
               return res.status(400).json({
--- a/tests/test_analyze.py
+++ b/tests/test_analyze.py
@@ -85,7 +85,13 @@ def test_interpret_prediction_returns_error_without_model():
 
 
 def test_predict_file_cli_outputs_json(tmp_path):
-    """Smoke test for the predict_file entrypoint used by the API."""
+    """Smoke test for the predict_file entrypoint used by the API.
+
+    Removes any cached model file first to simulate a cold-start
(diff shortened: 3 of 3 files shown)


ionfwsrijan:
@gopaljilab You may review and merge this

PR #456, opened 2026-05-31
fix: add periodic cleanup to rate limiter to prevent memory leak

### LINKED ISSUE #450: [BUG] Custom rate limiter has unbounded memory growth — no cleanup of expired entries
**Describe the bug**

The custom `publicApiRateLimiter` middleware in `server/middleware/rateLimiter.ts` uses a `Map<string, { count, resetTime }>` to track request counts per IP. When a window expires, the entry is overwritten for that IP on the next request — but entries for IPs that never return are **never deleted**. Over time, this Map grows unboundedly as unique IPs accumulate, causing a memory leak.

Additionally, the `x-forwarded-for` header can contain multiple comma-separated IPs, and using it directly as a Map key without parsing is unreliable.

**Affected file:** `server/middleware/rateLimiter.ts`

```typescript
const requestStore = new Map<string, { count: number; resetTime: number }>();

export const publicApiRateLimiter = (req: Request, res: Response, next: NextFunction) => {
  const ip = req.ip || req.headers["x-forwarded-for"] || "unknown";
  const now = Date.now();
  const record = requestStore.get(ip as string);

  if (!record || now > record.resetTime) {
    requestStore.set(ip as string, { count: 1, resetTime: now + WINDOW_MS });
    return next();
  }
  // ... entries are never cleaned up for IPs that don't return
};
```

**To Reproduce**

1. Start the server with `npm run dev`
2. Send requests from many unique IPs (or simulate with different `x-forwarded-for` headers)
3. Monitor Node.js heap memory usage over time
4. Observe that memory grows continuously as the Map accumulates stale entries that are never purged

**Expected behavior**

Expired rate limit entries should be periodically cleaned up to prevent unbounded memory growth. The Map should only contain entries for active or recently-active IPs.

**Screenshots**

N/A

**Desktop (please complete the following information):**

- OS: Any
- Browser: N/A (server-side issue)

**Additional context**

The project already uses the `express-rate-limit` library (for the assessment endpoint in `server/routes.ts`), which handles memory management internally. The custom rate limiter could either be replaced with `express-rate-limit` for consistency, or a cleanup interval should be added.

**Proposed fix:**
1. Add a periodic cleanup interval (e.g., every 5 minutes) that iterates the Map and deletes entries where `Date.now() > resetTime`
2. Alternatively, replace the custom implementation with the already-installed `express-rate-limit` library for consistency
3. Parse `x-forwarded-for` to extract only the first (client) IP

nyxsky404:
## Description

Fixes unbounded memory growth in the custom rate limiter middleware. The `requestStore` Map previously accumulated entries for every unique IP that ever made a request, but never cleaned up expired entries for IPs that didn't return. Over time this constitutes a memory leak that could crash the server.

## Related Issue

Fixes #450

## Type of Change

- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📝 Documentation update
- [ ] 🎨 Style / UI improvement
- [ ] ♻️ Refactor (no functional changes)
- [ ] ⚡ Performance improvement
- [ ] 🧪 Test addition or update
- [ ] 🔧 Chore / Tooling / Config

## Changes Made

- Added a `cleanupExpiredEntries()` function that iterates the Map and deletes entries where `Date.now() > resetTime`
- Scheduled cleanup to run every 5 minutes via `setInterval` with `.unref()` so it doesn't prevent graceful process exit
- Extracted a `getClientIp()` helper that properly parses `x-forwarded-for` headers (uses only the first/client IP instead of the raw header string)
- Removed unnecessary `as string` type casts

## Screenshots (if applicable)

N/A — server-side middleware fix.

## Testing

- [x] All existing tests pass

## Checklist

- [x] I have updated the documentation if needed

## Validation

Verified that the rate limiting behavior is preserved (same window, same max requests) while adding the cleanup mechanism. The `unref()` call ensures the timer doesn't interfere with graceful shutdown.

### DIFF
--- a/server/middleware/rateLimiter.ts
+++ b/server/middleware/rateLimiter.ts
@@ -3,14 +3,45 @@ import { Request, Response, NextFunction } from "express";
 const requestStore = new Map<string, { count: number; resetTime: number }>();
 const WINDOW_MS = 60 * 1000;
 const MAX_REQUESTS = 60;
+const CLEANUP_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes
+
+/**
+ * Periodically removes expired entries from the rate limit store
+ * to prevent unbounded memory growth from unique IPs that never return.
+ */
+function cleanupExpiredEntries() {
+  const now = Date.now();
+  for (const [key, record] of requestStore) {
+    if (now > record.resetTime) {
+      requestStore.delete(key);
+    }
+  }
+}
+
+// Schedule periodic cleanup to prevent memory leaks
+const cleanupTimer = setInterval(cleanupExpiredEntries, CLEANUP_INTERVAL_MS);
+cleanupTimer.unref(); // Allow the process to exit without waiting for this timer
+
+/**
+ * Extracts the client IP from the request, parsing x-forwarded-for
+ * to use only the first (original client) IP address.
+ */
+function getClientIp(req: Request): string {
+  const forwarded = req.headers["x-forwarded-for"];
+  if (typeof forwarded === "string") {
+    // x-forwarded-for can be comma-separated; first entry is the client
+    return forwarded.split(",")[0].trim();
+  }
+  return req.ip || "unknown";
+}
 
 export const publicApiRateLimiter = (req: Request, res: Response, next: NextFunction) => {
-  const ip = req.ip || req.headers["x-forwarded-for"] || "unknown";
+  const ip = getClientIp(req);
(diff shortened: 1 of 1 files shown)


nyxsky404:
Closing — this fix was already merged upstream in PR #407.

PR #520, opened 2026-06-01
fix: wire CSV injection sanitizer into export route (closes #510)

### LINKED ISSUE #510: [Security] Uploaded CSV patient data files are not sanitized, enabling CSV injection attacks
## Description

The bulk patient data upload feature accepts CSV files without sanitizing cell values. An attacker can craft a CSV file where cells start with `=`, `@`, `+`, or `-` characters (CSV injection / formula injection). When a clinician opens the exported report in Excel, arbitrary formulas execute.

## Exploit Scenario
1. Attacker submits a CSV with a field value: `=HYPERLINK("http://evil.com","Click me")`
2. The system processes and stores this value
3. A clinician exports data to Excel — the formula executes

## Suggested Fix
- Strip or escape leading formula characters from all CSV field values
- Prepend a tab character to neutralize formula injection

cc @gopaljilab

anuragbraveboy-sudo:
## Problem
PR #514 added `app/utils/csv_sanitizer.py` but never called it anywhere — the sanitizer was a dead module and the vulnerability remained open.  
 
This project uses a TypeScript/Node.js backend (not Python), so the fix needed to be implemented in TypeScript.

## What this PR does
- Adds `server/utils/csvSanitizer.ts` with three functions:
  - `sanitizeCell()` — neutralizes leading `=`, `+`, `-`, `@`, TAB, CR characters per OWASP CSV Injection guidelines
  - `sanitizeRow()` — sanitizes all values in a row object
  - `assessmentsToCsv()` — converts assessments to a fully sanitized CSV string

- Adds a new route `GET /api/assessments/export.csv` in `server/routes.ts` that:
  - Requires authentication
  - Fetches assessments from the database
  - Passes all values through the sanitizer before writing to CSV
  - Returns a safe downloadable CSV file

## Why this is better than #514
- Actually wired into the codebase — not just a utility file
- Correct language (TypeScript, matching the actual backend)
- Covers the export attack surface end to end

## References
- OWASP CSV Injection: https://owasp.org/www-community/attacks/CSV_Injection
- Issue #510


anuragbraveboy-sudo:
HEY @gopaljilab kindly check out this PR

anuragbraveboy-sudo:
Note: Please merge this PR first before the others, as subsequent PRs build on top of this branch.

gopaljilab (approved):
Hi @anuragbraveboy-sudo,

Thanks for putting this together! Implementing the CSV export route and ensuring it utilizes the assessmentsToCsv sanitizer is an important addition for securing our data exports against CSV injection.

Here are a few quick observations and suggestions before merging:

Hardcoded Pagination: On line 727, storage.getAssessments(1000, 0, userEmail) uses hardcoded pagination values (1000 limit, 0 offset). If a user has more than 1,000 assessments, their data will be truncated. Consider either lifting this limit entirely for exports or pulling the pagination/limit from the request query parameters (req.query).

Type Casting: On line 730, you use assessments as unknown as Record<string, unknown>[]. While this works to bypass strict compiler checks, it would be safer to properly type or cast the return shape of storage.getAssessments if possible, ensuring type safety downstream.

Error Handling Status: The 500 status code on failure is good, but you might want to ensure that no partial headers are written if an error occurs mid-stream (though since res.send(csv) happens at the end of the block, this is fine as-is).

Overall, the logic is clean, and wiring up requireAuth and requireVerified guarantees the endpoint is properly secured. Once we decide on handling the 1,000-record limit, this is good to go!

PR #521, opened 2026-06-01
fix: warm up ML model at startup to eliminate first-request latency (closes #509)

### LINKED ISSUE #509: [Performance] ML model is loaded from disk on every prediction request causing high latency
## Description

The trained ML model (pickle/joblib file) is loaded from disk on every incoming prediction request instead of being loaded once at application startup and cached in memory. This causes 2-5 second latency spikes per request.

## Impact
- Unacceptable latency for clinical use cases
- Disk I/O bottleneck under concurrent requests
- High CPU usage from repeated model deserialization

## Suggested Fix
- Load the model once at application startup (e.g., in Flask `app.py` before routes, or FastAPI lifespan event)
- Store the loaded model in a module-level variable or application context
- Add a model version check on startup

cc @gopaljilab

anuragbraveboy-sudo:
## Description
PR #513 added `app/ml/model_loader.py` (Python) but this project's backend is TypeScript/Node.js — that file was never called anywhere and the issue remained open.

This PR fixes the actual root cause: the ML model was being trained/loaded from disk on the very first prediction request, causing 2-5 second latency spikes.

## Related Issue
Closes #509

## Type of Change
- [x] ⚡ Performance improvement

## Changes Made
- Exported `getPythonExecutable()` from `server/routes.ts` so it can be reused at startup
- Added non-blocking ML model warmup in `server/index.ts` that runs `analyze.py train` immediately after the server boots
- Warmup uses `.then/.catch` so it never delays server startup
- If `diabetes_model.joblib` already exists and dataset is unchanged, warmup completes instantly
- First prediction request always finds a ready cached model — cold-start retraining eliminated

## Testing

## Checklist
- [x] My changes generate no new warnings


anuragbraveboy-sudo:
HEY @gopaljilab, kindly check out this PR.

anuragbraveboy-sudo:
Note: This PR should be merged after #510. The extra commits from the previous branch are inherited due to branch stacking — the actual change in this PR is only server/index.ts and server/routes.ts (model warmup).

PR #650, opened 2026-06-03
fix: canonical JSON stringify for deterministic request fingerprints

### LINKED ISSUE #573: [BUG] : generateRequestFingerprint uses JSON.stringify which produces non-deterministic fingerprints for logically identical payloads
## Description

In `server/routes.ts:25-32`, the `generateRequestFingerprint` function computes a SHA-256 hash from `JSON.stringify(payload)`. JavaScript's `JSON.stringify` does not guarantee property key ordering — the same logical object can serialize differently depending on the runtime, the order keys were assigned, or which code path parsed the input. Since the Zod schema that validates the request body may reorder or strip fields, and the client can send keys in any order, two identical clinical assessments can produce different fingerprints. This means the `activeInferenceRequests` deduplication set (line 631-636) never actually catches duplicate requests, and every submission goes through to the Python subprocess regardless.

The `userId` portion also uses `req.session.user?.email` (line 617) which can technically be undefined at the point the fingerprint is generated, producing a hash prefix of `"undefined::..."` for all unauthenticated scenarios, further weakening the dedup.

## Suggested fix

Use a canonical JSON serialization (e.g., `json-stable-stringify` or sort `Object.keys()` recursively before stringifying). Also add a non-null assertion or guard for the userId before computing the fingerprint.

ionfwsrijan:
## Description

Fixes `generateRequestFingerprint` using `JSON.stringify(payload)` which produces non-deterministic fingerprints for logically identical payloads due to JavaScript's unspecified property key ordering.

## Related Issue

Closes #573

## Type of Change

## Changes Made

- Removed duplicate function declaration in `server/routes.ts` (merge artifact where the function signature appeared twice)
- Added `canonicalStringify()` helper that recursively sorts `Object.keys()` before serializing, producing deterministic JSON output regardless of key insertion order, runtime, or code path
- Updated `generateRequestFingerprint` to use `canonicalStringify` instead of `JSON.stringify`
- Added `userId || "anonymous"` guard so the fingerprint prefix is never `"undefined::..."` for unauthenticated or edge-case scenarios

## Testing

- [x] All existing tests pass

## Checklist

### DIFF
--- a/server/routes.ts
+++ b/server/routes.ts
@@ -94,9 +94,22 @@ function parsePythonPrediction(stdout: string): PythonPrediction {
   return pythonPredictionSchema.parse(parsed);
 }
 
+function canonicalStringify(obj: unknown): string {
+  if (obj === null || typeof obj !== "object") {
+    return JSON.stringify(obj);
+  }
+  if (Array.isArray(obj)) {
+    return "[" + obj.map(canonicalStringify).join(",") + "]";
+  }
+  const keys = Object.keys(obj as Record<string, unknown>).sort();
+  const pairs = keys.map((k) => JSON.stringify(k) + ":" + canonicalStringify((obj as Record<string, unknown>)[k]));
+  return "{" + pairs.join(",") + "}";
+}
+
 function generateRequestFingerprint(payload: unknown, userId: string): string {
+  const uid = userId || "anonymous";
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #651, opened 2026-06-03
fix: remove misleading fake stats from calculateClinicalFallback (#574)

### LINKED ISSUE #574: [BUG] : calculateClinicalFallback returns statistically meaningless confidenceInterval and hardcoded modelConfidence that appear rigorous to clinicians
## Description

When the Python ML subprocess fails or times out, `server/routes.ts:456-533` runs a rule-based fallback (`calculateClinicalFallback`) that produces `confidenceInterval: riskScore ± 5` and hardcodes `modelConfidence: 0.95` for every single prediction. These values are displayed in the AssessmentResult UI alongside the risk score with the labels "95% CI" and "Model confidence". A clinician viewing the report has no way to tell these numbers are not real statistics — they look identical to the proper covariance-matrix-based confidence intervals produced by `analyze.py:324-352`. This defeats the purpose of having a fallback distinguish itself from a real ML result. The `disclaimer` footer does mention the fallback, but it's collapsed in the JSON response and easy to miss; the visual UI shows the same chart and confidence display regardless.

## Suggested fix

 Either omit confidenceInterval and modelConfidence entirely from the fallback response, or set them to null so the UI can show a "not available" indicator. Change the disclaimer to surface prominently in the UI rather than hiding it in a JSON field.

ionfwsrijan:
## Description

Fixes `calculateClinicalFallback` returning statistically meaningless `confidenceInterval: riskScore ± 5` and hardcoded `modelConfidence: 0.95` that appear identical to real covariance-matrix-based ML statistics, misleading clinicians viewing the report.

## Related Issue

Closes #574

## Type of Change

## Changes Made

- Removed `confidenceInterval` and `modelConfidence` from the `PredictionResult` interface in `server/routes.ts`
- Removed `confidenceInterval: riskScore ± 5` and hardcoded `modelConfidence: 0.95` from `calculateClinicalFallback` return value — these fields are now absent and correctly serialized as `null` in the JSON response
- Added `prediction.isFallback = isFallback` boolean to the prediction object so the frontend can conditionally surface the disclaimer prominently in the UI rather than hiding it in a JSON field
- Added `isFallback?: boolean` to the `AssessmentResponse` type in `shared/routes.ts`

## Testing

- [x] All existing tests pass

## Checklist

### DIFF
--- a/server/routes.ts
+++ b/server/routes.ts
@@ -229,8 +229,6 @@ interface PredictionResult {
   }>;
   clinicianAdvice: string[];
   patientAdvice: string[];
-  confidenceInterval?: string;
-  modelConfidence?: number;
 }
 
 function calculateClinicalFallback(input: unknown): PredictionResult {
@@ -378,8 +376,6 @@ function calculateClinicalFallback(input: unknown): PredictionResult {
           : [
             "Continue maintaining a healthy, balanced lifestyle and regular physical activity.",
           ],
-    confidenceInterval: `${Math.max(1, riskScore - 5)}% - ${Math.min(99, riskScore + 5)}%`,
-    modelConfidence: 0.95,
   };
 }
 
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #652, opened 2026-06-03
fix: log temp file cleanup failures instead of silently swallowing

### LINKED ISSUE #575: Temp files containing PII are silently left on disk when unlink fails in both preview and create assessment endpoints
## Description

Both `POST /api/assessments/preview` and `POST /api/assessments/create` write the patient's clinical data (age, BMI, HbA1c, blood glucose, smoking history, etc.) to a temp file at `os.tmpdir() + "/" + randomUUID() + ".json"`, pass the file path to the Python `analyze.py predict_file` subprocess, then attempt to delete the file in a `finally` block. However, the deletion is wrapped in a try/catch that silently swallows every error:

```typescript
finally {
  try { await unlink(tempFile); } catch (e) {}
}
```

On Windows, file locking by the Python subprocess can prevent immediate deletion. In Docker environments, `/tmp` is sometimes a shared volume with restrictive permissions. If unlink fails, the JSON file containing full PII remains on disk indefinitely. On a busy server with hundreds of assessments, these orphaned files accumulate and become a data leakage surface — anyone with filesystem access to the temp directory can read patients' clinical data with no authentication.

The `requestFingerprint` cleanup in the create endpoint (line 715-717) has the same silent-swallow pattern.

## Suggested fix

 Log the deletion failure (at least `console.warn`), and consider writing to a named temp directory that is wiped on server startup. Alternatively, pipe the JSON to Python's stdin instead of writing to disk at all — the `predict_file` command could accept input from stdin rather than a file path.

ionfwsrijan:
## Description

Fixes temp files containing patient PII (age, BMI, HbA1c, blood glucose, smoking history, etc.) being silently left on disk when `unlink` fails in both the preview and create assessment endpoints.

## Related Issue

Closes #575

## Type of Change

## Changes Made

- Preview endpoint (`server/routes.ts`): replaced silent `catch { // ignore }` with `catch (e) { console.warn("Failed to clean up temp file (preview):", tempFile, e); }`
- Create endpoint (`server/routes.ts`): replaced silent `catch { // ignore }` with `catch (e) { console.warn("Failed to clean up temp file (create):", tempFile, e); }`

Temp file deletion failures are now logged so operators can detect, investigate, and remediate PII data leakage on disk.

## Testing

- [x] All existing tests pass

## Checklist

### DIFF
--- a/server/routes.ts
+++ b/server/routes.ts
@@ -462,8 +462,8 @@ export async function registerRoutes(
       } finally {
         try {
           await unlink(tempFilePath);
-        } catch {
-          // ignore
+        } catch (e) {
+          console.warn("Failed to clean up temp file (preview):", tempFilePath, e);
         }
       }
     }
@@ -560,8 +560,8 @@ export async function registerRoutes(
         if (tempFilePath) {
           try {
             await unlink(tempFilePath);
-          } catch {
-            // ignore
+          } catch (e) {
+            console.warn("Failed to clean up temp file (create):", tempFilePath, e);
           }
         }
         if (requestFingerprint) {


ionfwsrijan:

PR #800, opened 2026-06-04
fix(tsc): resolve all TypeScript compilation and logger.error signatu…

Pratyush-Panda-2006:
Cleaned Monolithic Router (
server/routes.ts
): Removed redundant inline route definitions (bulk processing, exports, analytics, and detail handlers) that had already been decomposed into the modular server/routes/ files. Only the initialization logic, SimpleSemaphore class, and admin endpoints remain.
Fixed Pino Logger Signatures (
server/auth.ts
 & 
): Replaced invalid parameter patterns logger.error("Message", err) with logger.error({ err }, "Message") to match pino's strict type expectations.
Corrected Redis Connections & Worker Variables (
server/queue.ts
): Cast the redisConnection configuration as any to prevent type mismatches with BullMQ's inner dependencies and added userId to the destructured job.data values.


gopaljilab:
Thanks for putting this together, @Pratyush-Panda-2006! Great work cleaning up the monolithic router and fixing those Pino logger signatures.

It looks like some recent changes to main have introduced merge conflicts in server/queue.ts and server/routes.ts. Could you please rebase or merge main into your branch and resolve those conflicts so we can get this tested and merged? Let me know if you need any help with it!
And **make sure to mention which issue this PR resolved!**

PR #852, opened 2026-06-05
fix: use actual newline in assessmentsToCsv row join

### LINKED ISSUE #848: [BUG] Wrong newline escape in assessmentsToCsv (server/utils/csvExport.ts)
## Describe the bug
The rows are joined with "\\n" (a literal backslash-n string) instead of "\n" (an actual newline character). This means the exported CSV comes out as one single line with \n text between rows instead of proper line breaks — completely unreadable in Excel or any CSV parser.

## BUG: \\n is a literal two-character string "\n", not a newline

return [headers.join(","), ...rows].join( "\\\n");

The Fix:

## ✅ FIX: use \n for real newlines
return [headers.join(","), ...rows].join("\n");

maheshbhojane1:
Closes #848 

## Description

Fixes a broken CSV export in assessmentsToCsv where rows were joined with a literal two-character string \n (escaped backslash + n) instead of an actual newline character. This caused the entire export to be rendered as a single unbroken line, making the CSV completely unreadable in Excel, Google Sheets, or any standard CSV parser.

## Root Cause

In server/utils/csvExport.ts, the assessmentsToCsv function joins rows using "\\n" — a JavaScript string literal that produces the two characters \ and n, not a newline. Every CSV row was separated by the visible text \n rather than an actual line break.

// ❌ Before — literal backslash-n, not a newline
return [headers.join(","), ...rows].join("\\\n");

// ✅ After — real newline character
return [headers.join(","), ...rows].join("\n");

### DIFF
--- a/server/utils/csvExport.ts
+++ b/server/utils/csvExport.ts
@@ -2,5 +2,5 @@ export function assessmentsToCsv(data: Record<string, unknown>[]): string {
   if (data.length === 0) return "";
   const headers = Object.keys(data[0]);
   const rows = data.map(row => headers.map(h => JSON.stringify(row[h] ?? "")).join(","));
-  return [headers.join(","), ...rows].join("\\n");
+  return [headers.join(","), ...rows].join("\n");
 }


maheshbhojane1:
Hi @gopaljilab — PR is ready for review! It's a one-line fix for a broken CSV export (rows joined with literal \n instead of an actual newline). Could you approve the workflows to kick off CI? Happy to make any changes needed.

gopaljilab:
Awesome catch, @maheshbhojane1! Replacing that literal \n string with an actual newline character is exactly what's needed to fix the broken CSV formatting.

It looks like a recent merge to main has introduced a merge conflict in server/utils/csvExport.ts. Could you please update your branch against main and resolve the conflict so we can get this merged? Thanks!

PR #854, opened 2026-06-05
fix: use single quote instead of tab to neutralize CSV injection

### LINKED ISSUE #853: [BUG] CSV sanitizer neutralizes formula injection with a tab — but tab is itself a dangerous prefix (app/utils/csv_sanitizer.py)
## Describe the bug

sanitize_csv_value() prepends \t (tab) to neutralize dangerous formula characters like =, +, -, @. But \t is listed in DANGEROUS_PREFIXES itself! So any value starting with \t (a real tab character in data) is also prepended with another \t, producing \t\t<value> — corrupting the data. Worse, the comment even says "Prepend tab to neutralize" but the docstring says "Prepends a single quote" — a copy-paste inconsistency revealing this was changed hastily.

## DANGEROUS_PREFIXES = ('=', '+', '-', '@', '\t', '\r', '\n')  # \t is in here

def sanitize_csv_value(value):
    if s and s[0] in DANGEROUS_PREFIXES:
        return "\t" + s   # ❌ prepending \t, which is itself a DANGEROUS prefix!

def sanitize_csv_value(value: Any) -> str:
    if value is None:
        return ""
    s = str(value).strip()
    if s and s[0] in DANGEROUS_PREFIXES:
        return "'" + s   # ✅ single quote — safe, OWASP-recommended, not in DANGEROUS_PREFIXES
    return s

maheshbhojane1:
Closes #853 

## Description

Fixes a self-defeating CSV injection sanitizer in app/utils/csv_sanitizer.py, where the neutralizing prefix used to escape dangerous formula characters was \t (tab) — a character that is itself listed as dangerous in DANGEROUS_PREFIXES.
This caused any cell value starting with a real tab character to be double-escaped (\t\t<value>), corrupting the data. It also meant the OWASP mitigation was incomplete, since a crafted value beginning with \t followed by =, +, etc. could still be interpreted as a formula by spreadsheet applications. The fix replaces the tab prefix with a single quote ', which is the mitigation
explicitly recommended by OWASP and already described in the function's own docstring.

## Root Cause

In app/utils/csv_sanitizer.py, sanitize_csv_value prepends \t to neutralize dangerous formula characters. However \t is included in DANGEROUS_PREFIXES itself, creating two problems:

Data corruption — any legitimate cell value starting with a tab is double-escaped to \t\t<value>. Broken mitigation — the function's own docstring says it should prepend a single quote (OWASP recommendation) but the implementation uses tab instead, indicating a hasty edit that was never reconciled.

### DIFF
--- a/app/utils/csv_sanitizer.py
+++ b/app/utils/csv_sanitizer.py
@@ -33,8 +33,8 @@ def sanitize_csv_value(value: Any) -> str:
     s = str(value).strip()
 
     if s and s[0] in DANGEROUS_PREFIXES:
-        # Prepend tab to neutralize the formula
-        return "	" + s
+        # Prepend single quote to neutralize the formula (OWASP recommended)
+        return "'" + s
 
     return s
 


maheshbhojane1:
Hi @gopaljilab — PR is ready for review! Fixes a self-defeating CSV injection sanitizer where the neutralizing prefix \t was itself listed as dangerous. Replaced with ' per OWASP recommendation. Could you approve the workflows?

PR #878, opened 2026-06-05
Feat/search pagination metadata

### LINKED ISSUE #715: [FEATURE] Return total count and hasMore flag from search API for proper pagination
Is your feature request related to a problem? Please describe.
The search endpoints (/api/assessments/search) return only a raw array of results with no metadata. There is no way for the frontend to know how many total records match the query, whether more pages exist, or when to hide/show a "Load More" control. This makes it impossible to implement correct pagination UI without over-fetching.
Describe the solution you'd like
Wrap the search response in a metadata envelope:
json{
  "data": [...],
  "total": 84,
  "page": 1,
  "limit": 20,
  "hasMore": true
}
Add a countAssessments() method to IStorage and DatabaseStorage in server/storage.ts that mirrors the same filter conditions as searchAssessments but runs a SELECT count(*) query. Call it in the route handler alongside the existing data fetch and include the result in the response.
Describe alternatives you've considered

Cursor-based pagination (as proposed in #663/#659): More scalable for very large datasets but is a larger architectural change. The envelope approach here is additive and non-breaking — existing consumers that only read .data are unaffected, and cursor pagination can be layered on top later.
X-Total-Count response header: A valid HTTP convention but harder to consume in typed frontend code compared to a JSON body field.

Additional context
The page and limit query params are already parsed and validated in the route (searchValidation.ts). The only missing piece is surfacing the total back to the caller. This change touches server/storage.ts (new countAssessments method) and server/routes.ts (updated res.json call) — no schema migrations required.

OmanshiRaj:
## Description

This PR adds pagination metadata to the assessment search API response, allowing the frontend to implement accurate pagination controls without over-fetching.

Previously, /api/assessments/search returned only a raw array of results, making it impossible to determine the total number of matching records or whether additional pages existed.

### Changes Made

- Added a new countAssessments() method to IStorage.
- Implemented countAssessments() in DatabaseStorage.
- Reused the same filtering logic as searchAssessments() to ensure accurate counts.
- Updated the search endpoint to return a metadata envelope containing pagination information.
- Added hasMore calculation based on total records and current page.

### New Response Format

json {   "data": [...],   "total": 84,   "page": 1,   "limit": 20,   "hasMore": true } 

### Benefits

- Enables accurate pagination UI.
- Eliminates frontend guesswork and over-fetching.
- Provides total result counts for improved user experience.
- Maintains a clean foundation for future cursor-based pagination.

### Testing

- Verified total count matches search filters.
- Verified correct hasMore behavior across pages.
- Verified empty search results return total: 0.
- Confirmed paginated search functionality remains unchanged.

Closes #715


gopaljilab:
Great work on implementing the countAssessments tracking and updating the search endpoint, @OmanshiRaj!

We just have some minor merge conflicts in **server/routes.ts** and **server/storage.ts** due to recent updates on main. Once those are cleared up, we should be good to go!

PR #886, opened 2026-06-06
fix: prevent Express server from exiting immediately on startup (#786)

### LINKED ISSUE #786: [BUG] Express server exits immediately on startup due to misplaced process.exit(1)
Describe the bug
When attempting to start the server in development or production mode, the application immediately shuts down with exit code 1. The server fails to bind to the port or listen for incoming requests because of a misplaced statement.

Technical DetailsAt the very end of server/index.ts, the following block was appended:

    process.exit(1);

Since this statement is placed in the global scope of the module (outside the main asynchronous IIFE block (async () => { ... })()), Node.js evaluates it synchronously as soon as the module is imported, terminating the process immediately.

Steps to reproduce
Run npm run dev or node server/index.ts.
Notice that the command line exits instantly with error code 1 without showing the usual server listener log.

Prince2301p:
Fixes #786

### Changes Made

- Removed misplaced `process.exit(1)` that caused the Express server to terminate immediately after startup.
- Ensured the server continues running after successful initialization.
- Kept process termination logic only for genuine startup failures.

### Testing

- Installed dependencies.
- Started the application locally.
- Verified that the server remains active after startup.
- Confirmed no unexpected process termination occurs.

### Impact

Improves application reliability by allowing the Express server to stay running after successful startup.

### DIFF
--- a/client/src/components/CardioGuardLanding.tsx
+++ b/client/src/components/CardioGuardLanding.tsx
@@ -1,0 +1,427 @@
+import { useState } from "react";
+import { motion } from "framer-motion";
+import {
+  Activity,
+  ArrowRight,
+  CheckCircle2,
+  Github,
+  HeartPulse,
+  LineChart,
+  Linkedin,
+  LockKeyhole,
+  Mail,
+  PlayCircle,
+  ShieldCheck,
+  Stethoscope,
+  Workflow,
+} from "lucide-react";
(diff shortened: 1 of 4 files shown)


Prince2301p:
Hi @gopaljilab ,

I would like to work on this issue.

I will investigate the startup flow, identify the misplaced `process.exit(1)` call causing the Express server to terminate immediately after launch, and submit a fix with proper testing.

Thanks!

Issue #849, opened 2026-06-05
[BUG] Massive Route Duplication in server/routes.ts — 5x Database Queries Per Request


ionfwsrijan:
## Description
`server/routes.ts` (1556 lines) registers the exact same routes **5+ times each**. This means Express executes ALL matching handlers sequentially for every request:

- `/api/assessments/:id` (GET) — registered **5 times**
- `/api/assessments/search` — registered **5 times**
- `/api/assessments` (GET list) — registered **twice**
- `/api/assessments` (POST create) — registered **twice**
- `/api/assessments/preview` — registered **twice**
- `/api/assessments/:id` (DELETE) — registered **twice**

Each duplicate handler independently queries the database, logs access audit events, and attempts to send a response. Express sends only the first response, but the other 4 queries still execute and their results are discarded. This is a **5x amplification of database load** for every assessment API call.

Additionally, routes are split across `server/routes.ts`, `server/routes/assessments.routes.ts`, `server/routes/exports.routes.ts`, `server/routes/ml.routes.ts`, and `server/routes/analytics.routes.ts` with overlapping and inconsistent implementations.

## Files Affected
- `server/routes.ts` — all duplicate route registrations (entire file needs consolidation)
- `server/routes/assessments.routes.ts` — overlapping assessment routes
- `server/routes/exports.routes.ts` — overlapping export routes
- `server/routes/ml.routes.ts` — overlapping ML routes
- `server/routes/analytics.routes.ts` — overlapping analytics routes
- `shared/routes.ts` — route path definitions

## Expected Behavior
Each API route should be registered exactly once. A request to `/api/assessments/:id` should execute ONE handler that queries the database once and returns the response.

## Actual Behavior
Every assessment API route is registered between 2 to 5 times. Each duplicate handler independently:
1. Queries the database (same data 5 times)
2. Logs an access audit event (5 audit log entries per request)
3. Sends a response (only first one reaches client, 4 are discarded)
4. Runs middleware chains (5x middleware execution)

This causes 5x database load, 5x audit log volume, and 5x unnecessary processing for every assessment API call.

## Steps to Reproduce
1. Start the server and enable SQL query logging
2. Send a GET request to `/api/assessments/1`
3. Observe the server logs — the same SELECT query executes **5 times** against PostgreSQL
4. Check the audit log table — 5 identical access audit entries are created for the single request
[...]

PR #907, opened 2026-06-06
test: add route uniqueness assertion tests for all API routes

ionfwsrijan:
### Issue #849

Routes were being registered multiple times in `server/routes/assessments.routes.ts`, causing
`Cannot set headers after they are sent` errors. 

### Changes 
- Removed duplicate route registrations so each endpoint is registered exactly once
- Added `tests/routes.test.ts` — verifies `/api/assessments` route uniqueness at both
  test-setup and test-run levels

### Testing
```
npx jest tests/routes.test.ts
```


ionfwsrijan:

Issue #788, opened 2026-06-04
[BUG] storage.getAssessments returns cursor object but routing expects array


Pratyush-Panda-2006:
Describe the bug
The assessment listing endpoint throws runtime errors or sends malformed responses because it treats the return value of 
getAssessments
 as a raw array, whereas it actually returns a paginated object.

Technical Details
The method 
getAssessments
 in 
DatabaseStorage
 returns a promise of { data: Assessment[]; nextCursor: number | null }:
typescript

return { data: pagedData, nextCursor };
However, the calling endpoint in 
server/routes.ts
 attempts to read .length and slice properties directly off the returned value:
typescript

const assessments = await storage.getAssessments(limit, cursor, userEmail);
const nextCursor = assessments.length === limit && assessments.length > 0 
  ? assessments[assessments.length - 1].id 
  : undefined;
res.json({ data: assessments, nextCursor });
Since assessments is an object, assessments.length resolves to undefined, leading to runtime logical errors or client crashes. Similarly, 
server/routes/exports.routes.ts
 passes this returned object directly into CSV conversion helpers that expect arrays.

Prince2301p:
Hi,
I would like to work on resolving this issue. Please assign it to me. I’m a participant in GSSoC 2026 and would be happy to contribute.
Looking forward to collaborating with you.

PR #953, opened 2026-06-06
fix: handle paginated object returned by storage.getAssessments in ro…

Pratyush-Panda-2006:
issue fixed - #788 

Changes Made
Routing Endpoints Updated:

In 
server/routes.ts
: Updated the assessment listing endpoint to return the paginated object result directly, and modified the CSV export endpoint to map over the inner array assessments.data instead of the raw object.
In 
: Refactored the list endpoint to use cursor-based pagination (consistent with other list endpoints) and correctly return the paginated object.
In 
server/routes/exports.routes.ts
: Extracted and passed assessments.data into assessmentsToCsv(...).
In 
server/routes/patients.ts
: Changed the cursor argument from 0 to undefined to ensure it successfully fetches the first page of results.
Unit Tests Verified & Fixed:

In 
tests/routes.test.ts
:
Updated mockGetAssessments mock values to match the correct cursor-based pagination schema.
Updated the default export setup in the express-rate-limit mock to prevent import mismatch errors in modular routes.
Added emailVerified: true to the authenticated mock user session, satisfying the email verification gate middleware.
Modified assertions to verify the presence of nextCursor instead of the outdated total/page/totalPages.


gopaljilab:
Hey @Pratyush-Panda-2006, thanks for tackling this issue and updating the routing endpoints for proper cursor-based pagination! The automated checks are passing smoothly.

However, it looks like a merge conflict has popped up in server/routes/assessments.routes.ts. Could you please resolve it so we can move forward with getting your changes merged?

PR #1019, opened 2026-06-07
fix: propagate error dict from interpret_predictions_batch in single-record wrapper

### LINKED ISSUE #979: bug: interpret_predictions_batch in analyze.py crashes with UnboundLocalError (introduced in #871)
## Bug

Commit `9ab19325` (PR #871 — "perf(ml): vectorize batch predictions") introduced a broken `interpret_predictions_batch` function in `analyze.py`.

**Root cause:** The function was written with leftover code from the old single-record `interpret_prediction` at the top — it references `input_data` (undefined in this scope; the parameter is `input_data_list`) and uses `n_samples`/`n_features` before they are defined, causing an `UnboundLocalError` on every call.

**Affected line:** `analyze.py:338` — `cached = cache.get(input_data)`

**Fix:** Remove the stale preamble and initialize dimensions before the NumPy pre-allocation:
```python
n_samples = len(input_data_list)
n_features = len(features)
X_input = np.zeros((n_samples, n_features))
```

This fix has been applied to PRs #930 and #933 to unblock their CI. The bug exists on `main` and will affect any future branch that merges from `main`.

nyxsky404:
## Description

`interpret_predictions_batch` returns an error dict (`{"error": "..."}`) when the model is `None`, but the `interpret_prediction` single-record wrapper was calling `res[0]` unconditionally. Indexing a dict with `[0]` raises a `KeyError` instead of surfacing the original error, so callers never see the meaningful error message.

This PR adds a 2-line guard that detects the error dict and returns it directly, matching the existing contract expected by callers.

## Related Issue

Closes #979

## Type of Change

## Changes Made

- Added `isinstance(res, dict) and "error" in res` check in `interpret_prediction` before indexing `res[0]`, so the error dict from `interpret_predictions_batch` is propagated correctly to the caller.

## Testing

- [x] All existing tests pass

## Checklist

### DIFF
--- a/analyze.py
+++ b/analyze.py
@@ -544,6 +544,8 @@ def interpret_predictions_batch(model, scaler, features, input_data_list, cov_be
 def interpret_prediction(model, scaler, features, input_data, cov_beta=None):
     """Interprets a single patient's data, yielding clinician and patient views."""
     res = interpret_predictions_batch(model, scaler, features, [input_data], cov_beta)
+    if isinstance(res, dict) and "error" in res:
+        return res
     return res[0]
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


nyxsky404:
Hi @gopaljilab !, I have opened this pull request as a GSSoC '26 contributor and would appreciate your review when you get a chance.

PR #1037, opened 2026-06-07
fix: resolve temp file cleanup variable mismatch (#749)

### LINKED ISSUE #749: [BUG] : `ReferenceError` in Temp File Cleanup from Variable Name Mismatch (`tempFile` vs `tempFilePath`)
## Summary

The assessment creation handler declares a variable `tempFile` but the `finally` cleanup block references `tempFilePath` (never declared in this scope). Every successful assessment creation throws a `ReferenceError` after the response is already sent to the client. The temp file containing patient data is never deleted, leaving PII on disk.

---

## Root Cause

In `server/routes.ts`, the handler declares:

```typescript
let tempFile: string | null = null;   // declared as "tempFile"

try {
    tempFile = path.join(os.tmpdir(), `${randomUUID()}.json`);
    // ... write assessment data, run prediction ...
    return res.status(201).json({ ...assessment, prediction });
} catch (err) {
    // ...
} finally {
    if (tempFile) {                    // checks "tempFile" (correct)
        try {
            await unlink(tempFilePath); // BUG: uses "tempFilePath" (undefined!)
        } catch (e) {
            console.warn("Failed to clean up temp file (create):", tempFilePath, e);
        }
    }
}
```

The variable is declared as `tempFile` but the `finally` block references `tempFilePath` — a different identifier that is never declared in this scope. This is a copy-paste naming inconsistency.

Flow:
1. Request succeeds → `res.status(201).json()` sends response to client
2. `finally` block runs → `tempFile` is truthy, enters the `if` guard
3. `await unlink(tempFilePath)` → `tempFilePath` is `undefined` → throws `ReferenceError`
4. The temp file at `tempFile` is never deleted → PII remains on disk

---

## Impact

- Every successful assessment creation crashes the route handler after the client already received a 201 response.
- The temp file (containing patient assessment data) is leaked on disk — never cleaned up.
- `ReferenceError` may crash the Node.js process depending on unhandled rejection configuration.
- The `catch (e)` in the `finally` block catches the wrong thing — it was meant for `unlink` failures but now catches `ReferenceError` from the undefined variable.

---

## Suggested Fix

Rename the declaration to match the cleanup code, changing line:

```typescript
// BEFORE:
let tempFile: string | null = null;
// AFTER:
let tempFilePath: string | null = null;
```

And update the assignment:

```typescript
// BEFORE:
tempFile = path.join(...);
// AFTER:
tempFilePath = path.join(...);
```

This is a two-line change in one file. The `finally` block's `unlink(tempFilePath)` will then correctly reference the declared variable.

Prince2301p:
Fixes #749

Resolved a ReferenceError caused by using `tempFile` instead of `tempFilePath` during temporary file cleanup.

## Changes Made

- Replaced incorrect `tempFile` references with `tempFilePath`.
- Added a null check before cleanup.
- Updated warning logs to use the correct variable.

## Testing

- Verified the project builds successfully.
- Confirmed cleanup logic executes without ReferenceError.


Prince2301p:
Hi @gopaljilab ,

Implemented the fix for Issue #749 by replacing the incorrect tempFile reference with tempFilePath in the temporary file cleanup logic.

Changes Made
Fixed variable name mismatch in cleanup code.
Added a null check before attempting file deletion.
Updated cleanup logging to use the correct variable.
Verification
Confirmed the project passes TypeScript checks locally (npm run check).
Verified the cleanup logic no longer references an undefined variable.

Fixes #749

Thank you for your review.

gopaljilab:
Hi @Prince2301p,

Thanks for putting together this fix for the temporary file cleanup issue!

Before we can get this merged, there are a couple of things that need to be cleared up:

* **Merge Conflict:** There is a conflict in `server/routes.ts`. Could you please rebase with `main` or use GitHub's web editor to resolve it?
* **Failing Tests:** It looks like there are 3 failing CI checks right now (Python/pytest and TypeScript/Vitest). Could you take a look at the logs to see if the changes inadvertently broke any existing tests or expected data flows?

Let me know if you need any help with getting these sorted out!

ramnnn2006 (approved):

PR #1048, opened 2026-06-07
fix: resolve state desynchronization and data leakage in persistent ML daemon

### LINKED ISSUE #960: [BUG] State Desynchronization and Cross-Request Data Leakage in Persistent Python ML Daemon
Description
In production and development environments, the Express server handles machine learning inference by communicating with a persistent Python daemon process (analyze.py daemon) via standard input and output streams. The Python daemon processes requests sequentially in a single-threaded loop (for line in sys.stdin).

If a client request times out or is aborted on the Node.js side, the promise is rejected and the request is removed from the this.queue FIFO array. However, the background Python process is still executing that prediction. Once the Python process completes the calculation and writes the result to stdout, the daemon's line reader event (this.rl.on("line")) fires. It shifts the next active promise in this.queue and resolves it with the old, timed-out request's data.

This causes a severe race condition and cross-request state leakage: subsequent users will receive clinical prediction results belonging to a different patient.

Code Locations
server/services/mlService.ts
analyze.py
Steps to Reproduce
Trigger an ML inference request that takes longer than ML_TIMEOUT_MS (e.g., inject a delay in analyze.py or overload the CPU).
The Node.js request times out, rejecting the client's request and removing it from the this.queue array.
Submit a new, valid request immediately after.
When the Python daemon finishes processing the first request, it outputs the result to stdout.
The Node.js daemon listener receives the output, performs this.queue.shift(), and incorrectly resolves the second request's promise with the first patient's data.

Pratyush-Panda-2006:
Fixes #960

### Description
This PR fixes state desynchronization and cross-request data leakage issues in the persistent Python ML daemon. Previously, if a request timed out or was aborted, the daemon process would still finish the computation and return stdout results which would then be incorrectly shifted and mapped to the next active request in a FIFO queue. This led to cross-request data leakage.

### Changes Made
- Updated `analyze.py` to support a `daemon` mode that reads inputs from stdin and outputs results to stdout tagged with a unique `requestId`.
- Refactored `server/services/mlService.ts` to use a persistent `PythonDaemonManager` via `spawn`, maintaining active requests in a `Map<string, PendingRequest>` keyed by `requestId`. This avoids any FIFO-based desynchronization and ensures requests and responses are strictly paired using UUIDs.
- Refactored the route handlers for `POST /api/assessments/preview` and `POST /api/assessments/simulate` in `server/routes.ts` to use `MLService.runAssessmentInference(input)` instead of launching one-off python executions, ensuring they run through the persistent daemon.
- Updated unit tests in `tests/routes.test.ts` to spy on the `pythonDaemon.predict` method on a test-by-test basis, preventing test cross-pollution and validating correct timeout behavior (returning 503 for preview and 408 for simulate).


gopaljilab:
Hi @Pratyush-Panda-2006,

Thanks for putting this PR together to resolve the state desynchronization and data leakage issues!

Before this can be merged, we just need to address a couple of things:

* **Merge Conflict:** There is a conflict in `analyze.py`. Could you please rebase with `main` or use GitHub's web editor to resolve it?
* **Failing Tests:** The [[TypeScript and Vitest Tests](https://github.com/gopaljilab/Clinical-Insight-Engine/pull/1048)](https://github.com/gopaljilab/Clinical-Insight-Engine/pull/1048) check is currently failing. Could you take a look at the logs to see what broke?

Let me know if you need any help getting this sorted out!

Issue #1158, opened 2026-06-11
[FEATURE] Add end-to-end request observability for assessment and ML workflows


RishiByte:
## Problem

The app has separate frontend, Express API, queue, and Python ML execution paths, but a single clinical assessment request is not traceable end-to-end. When an assessment stalls, falls back, fails validation, or gets queued, maintainers have to manually stitch together browser behavior, Express logs, queue state, and Python subprocess output.

This is especially risky for clinical workflows because failures can appear as generic "assessment failed" or fallback behavior without enough context to debug production incidents quickly.

## Proposed solution

Add request-scoped observability across the full assessment lifecycle:

- Generate or reuse a correlation/request ID at the API boundary.
- Propagate that ID through assessment creation, queue processing, ML subprocess execution, fallback handling, and response payload metadata where appropriate.
- Include structured log fields for request ID, user/session scope, route, assessment job ID, ML execution duration, fallback usage, timeout, and error class.
- Surface a support-safe request ID in frontend error states/toasts so users can report reproducible failures without exposing PHI.
- Add tests proving the ID is preserved across direct and queued assessment flows.

## Suggested implementation pointers

- `server/middleware/requestId.ts`
- `server/routes/assessments.routes.ts`
- `server/queue.ts`
- `server/services/mlService.ts`
- `server/logger.ts`
- `client/src/hooks/use-assessments.ts`
- Existing tests under `tests/` and `server/*.test.ts`

## Acceptance criteria

- Every assessment request has a stable correlation ID from entry to response.
- Direct inference and queued inference logs include the same request/job context.
- ML fallback and timeout logs include duration, fallback status, and non-PHI error metadata.
- Frontend assessment failures display or expose a support-safe reference ID.
- Tests cover correlation propagation for successful, failed, and queued assessment paths.
- No PHI is added to logs.

## Difficulty

Advanced

## Expected outcome

Maintainers can debug production assessment incidents across frontend, API, queue, and ML layers without relying on unstructured logs or PHI.

RishiByte:
@gopaljilab assign me Please add Gssoc tag in int

PR #1164, opened 2026-06-11
Propagate request IDs through assessment queue

RishiByte:
## Summary
- includes the API request ID in queued assessment responses
- passes requestId into assessment queue jobs
- adds structured queue logs for job start, ML completion, and failure paths
- keeps the queued assessment response contract aware of the optional requestId

## Related issue
Supports #1158

## Verification
- git diff --check

Note: focused test execution is still blocked locally because dependency installation fails in this checkout: npm ci reports package-lock drift around optional bufferutil, and npm install exits with npm's internal 'Exit handler never called' error.


RishiByte:

Issue #1159, opened 2026-06-11
[FEATURE] Add shared API contract tests for route schema drift


RishiByte:
## Problem

The codebase has shared route schemas and a centralized frontend API layer, but there is no full contract test layer proving that Express responses, shared route definitions, and React consumers stay aligned over time.

As the API grows, small response-shape changes can silently break frontend parsing, assessment history, exports, or authentication flows.

## Proposed solution

Build an API contract testing workflow around the shared route definitions:

- Generate or maintain a contract map from `shared/routes.ts`.
- Add server-side tests that verify key API responses match the declared Zod response schemas.
- Add client-side tests for `ApiClient` and assessment hooks that use representative success/error payloads.
- Add CI-friendly documentation describing how contributors should update contracts when changing routes.
- Include coverage for error responses, pagination metadata, and assessment creation responses.

## Suggested implementation pointers

- `shared/routes.ts`
- `shared/routes.test.ts`
- `client/src/lib/apiClient.ts`
- `client/src/hooks/use-assessments.ts`
- `server/routes/*.ts`
- `tests/routes.test.ts`
- `tests/auth.routes.test.ts`

## Acceptance criteria

- Contract tests fail when a route returns a payload that does not match its shared schema.
- Assessment list, assessment creation, auth, export, and ML-related routes have representative contract coverage.
- Error responses are covered, not only happy paths.
- Frontend tests verify `ApiClient` behavior for JSON, empty responses, and standardized error payloads.
- Documentation explains the workflow for updating route contracts.

## Difficulty

Advanced

## Expected outcome

Frontend and backend contributors can evolve APIs safely, with automated tests catching client/server schema drift before it reaches users.

PR #1165, opened 2026-06-11
Add shared API contract tests

RishiByte:
## Summary
- adds shared route contract tests for queued assessment responses
- verifies assessment list pagination metadata shape
- checks validation error responses reject malformed payloads

## Related issue
Supports #1159

## Verification
- git diff --check


RishiByte:

Issue #1160, opened 2026-06-11
[FEATURE] Add privacy-conscious assessment draft recovery


RishiByte:
## Problem

Assessment creation is a multi-field clinical workflow, but users can lose in-progress assessment data if the session expires, the network drops, the page reloads, or the queued assessment flow times out. This creates a poor experience and may cause clinicians to re-enter sensitive data from scratch.

Because assessment forms contain PHI, a naive local-storage autosave would also create privacy risk.

## Proposed solution

Add a privacy-conscious assessment draft recovery workflow:

- Autosave in-progress assessment input as a local draft with clear expiration.
- Avoid storing drafts indefinitely and never sync drafts to the server unless the user submits.
- Add a visible restore/discard flow when a draft is detected.
- Clear drafts after successful submission, logout, or expiration.
- Use safe storage practices and document the privacy tradeoffs.
- Add tests for draft creation, restore, discard, expiration, and successful-submit cleanup.

## Suggested implementation pointers

- `client/src/pages/Dashboard.tsx`
- `client/src/hooks/use-assessments.ts`
- `client/src/lib/apiClient.ts`
- `client/src/components/auth/AuthFlow.tsx`
- `client/src/components/ui/dialog.tsx`
- `client/src/components/ui/toast.tsx`
- `docs/data-privacy.md`

## Acceptance criteria

- In-progress assessment data can be restored after reload or temporary navigation.
- Drafts expire automatically and can be discarded manually.
- Drafts are cleared after successful assessment submission and logout.
- The UI clearly lets users restore or discard a detected draft.
- Tests cover restore, discard, expiration, and cleanup behavior.
- Privacy documentation explains where drafts are stored and when they are removed.

## Difficulty

Advanced

## Expected outcome

Clinical users get a safer, more resilient assessment workflow without introducing indefinite PHI persistence in the browser.

RishiByte:
@gopaljilab Please add Gssoc tag init

PR #1166, opened 2026-06-11
Add assessment draft storage utilities

RishiByte:
## Summary
- adds a versioned local assessment draft storage utility
- expires stale drafts automatically to avoid indefinite PHI persistence
- adds explicit draft clearing support for successful submit/logout flows
- covers restore, expiration cleanup, and manual clear behavior with tests

## Related issue
Supports #1160

## Verification
- git diff --cached --check


RishiByte:

Issue #1162, opened 2026-06-11
[FEATURE] Add golden clinical scenario tests for risk outputs and recommendations


RishiByte:
## Problem

Clinical risk outputs, fallback recommendations, biomarker alerts, prediction explanations, and recommendation text can change without a clear golden-test safety net. This makes it hard to distinguish intentional clinical behavior updates from accidental regressions.

For a clinical decision-support project, contributors need stronger safeguards around deterministic outputs and explanation quality.

## Proposed solution

Create a golden clinical scenario test suite:

- Define a small set of representative synthetic patient scenarios covering low, moderate, high, edge-case, and missing-data profiles.
- Snapshot expected risk category, important factors, fallback disclosure, recommendation themes, and explanation structure.
- Add tests for the rule-based fallback path and clinical helper services.
- Document how to update golden scenarios when clinical logic intentionally changes.
- Ensure scenarios are synthetic and contain no PHI.

## Suggested implementation pointers

- `server/services/mlService.ts`
- `server/services/recommendation-engine.ts`
- `server/services/prediction-explainer.ts`
- `server/services/biomarker-trend-analyzer.ts`
- `server/services/assessment-quality-checker.ts`
- `tests/`
- `docs/clinical-validation.md`

## Acceptance criteria

- A documented synthetic golden dataset exists for clinical behavior tests.
- Tests cover low, moderate, high, and edge-case patient profiles.
- Fallback risk scoring and recommendation output are covered.
- Prediction explanation structure is validated without overfitting to brittle wording.
- Documentation explains how maintainers should review intentional clinical-output changes.
- No real patient data is used.

## Difficulty

Advanced

## Expected outcome

Clinical behavior becomes safer to refactor because major changes to recommendations, explanations, and fallback risk scoring are visible in tests.

RishiByte:
@gopaljilab Please add Gssoc tag init

PR #1168, opened 2026-06-11
Add clinical fallback golden tests

RishiByte:
## Summary
- adds synthetic golden scenarios for the clinical fallback risk calculator
- covers stable low-risk, moderate prediabetes, and high-risk metabolic profiles
- asserts risk category, score, key factors, advice, and confidence interval behavior

## Related issue
Supports #1162

## Verification
- git diff --cached --check


RishiByte:

PR #1281, opened 2026-06-14
fix(what-if): resolve temporary file vulnerability and limit perturbations size in batch simulation

### LINKED ISSUE #1256: bug: Disk exhaustion vulnerability due to unbounded temporary JSON file generation in /what-if/batch
Description
The /what-if/batch endpoint in 
 creates a temporary JSON file on disk for every incoming request:

typescript

const tempFile = path.join(os.tmpdir(), `${randomUUID()}.json`);
...
await writeFile(tempFile, JSON.stringify(payload));
...
const stdout = await new Promise<string>((resolve, reject) => {
  const child = execFile(
    getPythonExecutable(),
    [analyzePyPath, "counterfactual", tempFile],
    ...
If the backend crashes mid-execution (or if the subprocess times out and throws an unhandled promise rejection before the finally block is executed), the file cleanup logic inside finally might be bypassed or skipped. Furthermore, there is no restriction on the size of the request body perturbations payload, which allows large files to be written to the server's temp directory.

Impact
Disk Exhaustion (DoS): A malicious user sending large batch requests or triggering process timeouts could saturate the /tmp or disk storage of the container, leading to service crashes.

Pratyush-Panda-2006:
Closes #1256

### Changes
1. **Removed Temporary File Writes**: Refactored the `/what-if/batch` route in `server/routes/assessments.routes.ts` to pass the JSON payload directly to the Python subprocess's `stdin` stream (rather than writing to a temporary file in `os.tmpdir()`). This completely eliminates disk operations and avoids any lingering file leaks if the backend crashes or times out.
2. **Enforced Input Constraints**: Added a `.max(50)` size constraint to the `perturbations` array in `shared/routes.ts` to restrict the maximum complexity of the input payload and prevent DoS / memory exhaustion.
3. **Fixed Upstream Route Test**: Updated the `accepts paginated assessment list metadata` test in `shared/routes.test.ts` to match the current cursor-based schema in the codebase (which only expects `data` and `nextCursor`), resolving an upstream test failure.
4. **Added Batch Simulation Tests**: Added new tests in `tests/routes.test.ts` covering both successful processing via standard input and validation failure when the number of perturbations exceeds 50.

### Verification
Ran the test suite locally:
```bash
npm run test


gopaljilab:
Hi @Pratyush-Panda-2006 , 

I've noticed a merge conflict in `shared/routes.ts` due to concurrent changes on the `main` branch. 

To resolve it and keep the security enhancements from this PR intact, we need to ensure that the `.max(50)` constraint on the `perturbations` array remains applied to the schema within `shared/routes.ts`. This constraint is crucial as it prevents DoS and memory exhaustion vulnerabilities during batch simulations.

I will pull the latest updates from `main` into this branch, manually merge the schema definitions to preserve the input validation limits, and push the resolution shortly.

Pratyush-Panda-2006:
@gopaljilab I have fixed the conflicts

Issue #1253, opened 2026-06-13
bug: Race condition/DoS in analyze.py lock cleanup logic when concurrent requests invoke _clean_stale_lock()


Pratyush-Panda-2006:
Description
The ML training pipeline uses a file-based locking mechanism (_acquire_lock, _release_lock, _clean_stale_lock) in 
analyze.py
 to protect the model file diabetes_model.pkl from concurrent writes during retraining. However, _clean_stale_lock() contains a race condition:

python

def _clean_stale_lock():
    try:
        mtime = os.path.getmtime(LOCK_FILE)
        if time.time() - mtime > 300:
            os.remove(LOCK_FILE)
    except OSError:
        pass

Under concurrent invocations, one thread/process can read the mtime, check the timeout condition, and yield control to another thread. The second thread deletes the lock file. The first thread then attempts os.remove(LOCK_FILE) which raises an OSError (caught silently). More critically, during a lock retry loop in _acquire_lock, there is a risk of active locks being erroneously deleted if system clock synchronization issues or filesystem delays skew mtime values.

Impact
Concurrency Failures: Concurrent requests triggering model retraining can corrupt diabetes_model.pkl or its signature file if one process deletes the lock of another active process.
Model Signature Failure: If the signature verification fails due to file corruption, the server throws a PermissionError and falls back entirely to rule-based algorithms, degrading diagnostics quality.

Pratyush-Panda-2006:
assign/ GSSOC

PR #1284, opened 2026-06-14
fix(ml-locking): resolve race condition in file locking mechanism

Pratyush-Panda-2006:
Issue #1253

I have successfully resolved the concurrency lock race condition in analyze.py (Issue 1253) and pushed the fix to the repository. 
 
Summary of Changes
Advisory OS-Level Locks: Transitioned from a custom timeout-based lock cleanup file mechanism to a standard advisory file locking scheme:
On Windows: Uses msvcrt.locking.
On Unix/Linux/macOS: Uses fcntl.flock.
Reentrancy and Exclusivity: Aligned the lock's acquire method to return False when a lock is already held by the current thread/instance, ensuring standard non-reentrant exclusivity behavior to satisfy the test_lock_is_exclusive unit test.
Automatic Cleanup: Since the OS-level advisory locks are automatically cleaned up when the process closes/terminates, _clean_stale_lock() was refactored into a safe no-op.


gopaljilab:
Hi @Pratyush-Panda-2006! Thanks for tackling the concurrency lock race condition in analyze.py and migrating to standard advisory OS-level file locking (msvcrt / fcntl)—this is a much cleaner approach to handling the exclusivity behavior.

I noticed that while the Python test suites passed successfully, the CI pipeline is currently failing on the frontend/TypeScript check:

![TypeScript and Vitest Tests / test (pull_request)] TypeScript and Vitest Tests / test Could you take a look at the Vitest logs to see if the changes to the file locking mechanism or the refactored _clean_stale_lock() no-op inadvertently broke an integration test or an expected state on the TypeScript side?

Let me know if you need any help digging through the logs!

PR #1404, opened 2026-06-16
fix(#1321): fix RLS bypass caused by separate AsyncLocalStorage instances

### LINKED ISSUE #1321: Row-Level Security (RLS) Completely Bypassed — All Repository Methods Use Global Connection Pool
## Description

The application implements sophisticated PostgreSQL Row-Level Security (RLS) policies in migration `0004_enable_rls.sql` with per-user session variables (`app.current_user_id`, `app.current_user_email`, `app.current_user_role`). However, **every single repository method** calls `getDb()` which returns the **global connection pool** instead of the per-request RLS client. The RLS middleware creates a dedicated client with session variables set, but no code path actually uses it. The entire multi-file RLS infrastructure is decorative — **no Row-Level Security is ever enforced**.

### The Problem

**assessment.repository.ts — ALL methods:** Every repository query uses the global pool:
```typescript
import { getDb } from "../db";

async function getAssessmentsByPatientName(patientName: string, limit: number, offset: number) {
  const db = getDb();  // ← Global pool with full database access
  return db.select().from(schema.assessments)
    .where(eq(schema.assessments.patientName, patientName))
    .limit(limit).offset(offset);
}

async function createAssessment(data: InsertAssessment) {
  const db = getDb();  // ← Global pool, NOT RLS client
  return db.insert(schema.assessments).values(data).returning();
}
```

**db-rls.ts (lines 16-18):** The RLS-aware database accessor is defined but never called by any repository:
```typescript
const rlsStorage = new AsyncLocalStorage<NodePgDatabase<typeof schema>>();

export function getRlsDb(): NodePgDatabase<typeof schema> | undefined {
  return rlsStorage.getStore();  // ← Returns undefined everywhere since nothing sets it
}
```

**rlsContext.ts (lines 49-119):** The middleware that sets up the RLS context:
```typescript
export async function rlsContextMiddleware(req: Request, res: Response, next: NextFunction) {
  const user = req.session?.user;
  if (user) {
    const rlsDb = createRlsClient(user.email, user.id, user.role);
    // Sets app.current_user_id, app.current_user_email, app.current_user_role
    // via SET LOCAL statements on this connection
    rlsStorage.run(rlsDb, () => {
      next();  // ← RLS client available in AsyncLocalStorage, but repos never fetch it
    });
  } else {
    next();
  }
}
```

**migrations/0004_enable_rls.sql:** RLS policies defined on the `assessments` table:
```sql
ALTER TABLE assessments ENABLE ROW LEVEL SECURITY;

CREATE POLICY user_assessments_policy ON assessments
  USING (created_by = current_setting('app.current_user_email')::text);
```

### Impact

1. **Zero access control enforcement**: Every `SELECT`, `INSERT`, `UPDATE`, and `DELETE` on the `assessments` table runs against the unrestricted global pool. The RLS policies that check `current_setting('app.current_user_email')` against `created_by` are never triggered.
2. **Any authenticated user can access ANY patient's data**: All assessment data (including PHI like patient names, risk scores, biomarkers) is readable by any clinician in the system.
3. **Data exfiltration risk**: A malicious insider or compromised account can query the entire assessments table without restriction.
4. **False sense of security**: The RLS migration file exists, the middleware sets up the context, code reviewers see the RLS infrastructure and assume it works — but it's fully decorative.
5. **Every repository is affected**: Not just assessments — any repository that uses `getDb()` instead of `getRlsDb()` has the same issue.

### Required Changes

1. **`db-rls.ts`**: Export the `rlsStorage` instance so middleware can use it, and ensure `getRlsDb()` is the primary accessor.

2. **`assessment.repository.ts` (and all other repositories)**: Change `getDb()` calls to use `getRlsDb()` with a fallback:
   ```typescript
   import { getRlsDb } from "../db-rls";
   import { getDb } from "../db";

   function getSecureDb() {
     return getRlsDb() ?? getDb();  // Use RLS client if available, fall back to global
   }
   ```
   Or, make the repository accept the db instance via constructor/dependency injection.

3. **`storage.ts`**: Either accept the db instance as a parameter on each method call, or refactor storage to use a request-scoped factory pattern.

4. **`middleware/rlsContext.ts`**: Ensure the middleware is registered in the correct order (after auth, before route handlers) in the Express middleware chain. Verify it runs for all API routes, not just a subset.

5. **Remove the global `getDb()` calls** from repositories entirely in production, or add a runtime check that logs a critical warning if `getRlsDb()` returns undefined in production.

6. **Add integration tests** that create two users, have each create assessments, and verify that user A cannot read user B's assessments through any API endpoint.

This is approximately **80-130 lines of changes across 4-5 files** — every repository and storage method needs to be RLS-aware, and the middleware setup needs to be verified.

ionfwsrijan:
## Summary

The entire RLS infrastructure was completely non-functional — all repository queries bypassed Row-Level Security because `db-rls.ts` and `db.ts` used **separate `AsyncLocalStorage` instances**, so the per-request RLS client set by the middleware was never found by `getDb()`.

### Root Cause

- `db.ts` (line 7): `export const dbRlsStorage = new AsyncLocalStorage(...)` — used by `getDb()` to check for a per-request RLS database client
- `db-rls.ts` (line 14): `const rlsStorage = new AsyncLocalStorage(...)` — used by `runWithRlsDb()` called from the `rlsContext` middleware

Since these were **two different `AsyncLocalStorage` instances**, the RLS database client set by the middleware (stored in `rlsStorage`) was **never found** by `getDb()` (which only checked `dbRlsStorage`). Every repository query fell through to the global connection pool, completely bypassing all RLS policies.

### Fix

| File | Change |
|---|---|
| `server/db-rls.ts` | Import and reuse `dbRlsStorage` from `db.ts` instead of creating a separate `AsyncLocalStorage` instance |

Now both `runWithRlsDb()` and `getDb()` read from and write to the **same** `AsyncLocalStorage`, making the full RLS chain functional:

1. `rlsContext` middleware creates a dedicated DB client with `SET LOCAL` session variables
2. Stores it via `dbRlsStorage.run(db, next)` 
3. Repository methods call `getDb()` which finds the RLS client in `dbRlsStorage`
4. PostgreSQL RLS policies are triggered via the session variables

Closes #1321

### DIFF
--- a/server/db-rls.ts
+++ b/server/db-rls.ts
@@ -1,8 +1,7 @@
-import { AsyncLocalStorage } from "async_hooks";
 import { drizzle, type NodePgDatabase } from "drizzle-orm/node-postgres";
 import pg from "pg";
 import * as schema from "@shared/schema";
-import { getPool } from "./db";
+import { getPool, dbRlsStorage } from "./db";
 
 export interface RlsUserContext {
   userId: string;
@@ -11,10 +10,8 @@ export interface RlsUserContext {
   patientName?: string;
 }
 
-const rlsStorage = new AsyncLocalStorage<NodePgDatabase<typeof schema>>();
-
 export function getRlsDb(): NodePgDatabase<typeof schema> | undefined {
-  return rlsStorage.getStore();
+  return dbRlsStorage.getStore();
 }
 
 export async function createRlsClient(context: RlsUserContext): Promise<{
@@ -56,5 +53,5 @@ export function runWithRlsDb<T>(
   db: NodePgDatabase<typeof schema>,
   fn: () => T,
 ): T {
-  return rlsStorage.run(db, fn);
+  return dbRlsStorage.run(db, fn);
 }


ionfwsrijan:

ionfwsrijan:
@gopaljilab

PR #1416, opened 2026-06-17
[FIX] Hash password reset tokens before storing in database

### LINKED ISSUE #1411: [BUG] Security: Password reset tokens stored in plaintext in database — no hashing
## Summary

Password reset tokens are generated as hex strings and stored **directly in plaintext** in the \password_reset_tokens\ database table. If the database is ever breached (SQL injection, backup leak, DB admin compromise), all active reset tokens are exposed, allowing an attacker to reset any user's password without knowing their credentials.

## File

\server/auth.ts\ (lines 819-827)

## The Problematic Code

\\\	ypescript
const token = randomBytes(32).toString('hex');
const expiresAt = new Date(Date.now() + 60 * 60 * 1000);

await db.insert(passwordResetTokens).values({
    userId: user.id,
    token,           // <-- plaintext token stored in DB
    expiresAt,
    used: false,
});
\\\

Later, when the user clicks the reset link, the token is looked up directly:

\\\	ypescript
const result = await db
  .select()
  .from(passwordResetTokens)
  .where(eq(passwordResetTokens.token, token))  // <-- compared in plaintext
  .limit(1);
\\\

## Why This Is a Problem

1. **Database breach = all tokens compromised** — Unlike password hashes (which use bcrypt), these tokens are stored as raw hex strings.
2. **Long-lived tokens** — The token has a 1-hour TTL (\60 * 60 * 1000\ = 1 hour), giving a wide window for exploitation.
3. **No rate limiting on token redemption** — An attacker with a leaked database can attempt all tokens offline.
4. **Industry best practice violation** — OWASP and auth libraries (like Django, Devise) always hash reset tokens before storage.

## Required Fix (~60 lines change)

1. Hash the token before storing it (using SHA-256):
   \\\	ypescript
   import { createHash } from 'crypto';
   const tokenHash = createHash('sha256').update(token).digest('hex');
   // Store tokenHash instead of token
   \\\
2. When looking up the token on reset, hash the incoming token before querying:
   \\\	ypescript
   const incomingHash = createHash('sha256').update(incomingToken).digest('hex');
   const result = await db
     .select()
     .from(passwordResetTokens)
     .where(eq(passwordResetTokens.token, incomingHash))
     .limit(1);
   \\\
3. Update the database schema/column name if needed (e.g., \	oken_hash\ instead of \	oken\)
4. Add a comment explaining why the token is hashed (defense-in-depth, not because the token is a secret — it IS a secret)
5. Verify the \used\ flag and \expiresAt\ checks still work correctly

**Note**: This does NOT change the user-facing flow — the reset link URL still contains the raw token. Only the database representation changes.

sanrishi:
Closes #1411

### DIFF
--- a/server/auth.ts
+++ b/server/auth.ts
@@ -1,5 +1,5 @@
 import { Router, type Request, type Response, type NextFunction } from "express";
-import { randomInt, randomBytes } from "crypto";
+import { randomInt, randomBytes, createHash } from "crypto";
 import bcrypt from "bcrypt";
 import { rateLimit } from "express-rate-limit";
 import { issueToken } from "./services/auth/tokenValidator";
@@ -639,9 +639,19 @@ out
       }
 
       const token = randomBytes(32).toString("hex");
+      const tokenHash = createHash("sha256").update(token).digest("hex");
       const expiresAt = new Date(Date.now() + 60 * 60 * 1000);
 
+<<<<<<< HEAD
       await authRepository.createPasswordResetToken(user.id, token, expiresAt);
+=======
+      await db.insert(passwordResetTokens).values({
+        userId: user.id,
+        token: tokenHash,
+        expiresAt,
+        used: false,
+      });
+>>>>>>> b9563ad5 (fix(security): hash password reset tokens before storing in database)
 
       const resetLink = `${process.env.APP_URL || "http://localhost:5173"}/reset-password?token=${token}`;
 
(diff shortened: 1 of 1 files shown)


sanrishi:

gopaljilab:
Thanks for the contribution, @sanrishi! It looks like this PR has run into quite a few merge conflicts with the main branch across several client and server files. Could you please sync your branch with main and resolve the conflicts?

Once that's done, I'll authorize the Vercel deployment so we can check out the preview build and let the CI pipeline finish running.

sanrishi:
@gopaljilab Opened fresh prs please review!

sanrishi:

PR #1417, opened 2026-06-17
[FIX] Make FileLock reentrant and fix TOCTOU race in Python model loading

### LINKED ISSUE #1415: [BUG] Race condition: Non-reentrant FileLock and TOCTOU in Python model loading (analyze.py)
## Summary

The `analyze.py` module has two distinct concurrency bugs in its file-locking and model-loading code:

1. **Non-reentrant FileLock** — The lock silently returns `False` on reentrant calls from the same thread, which can cause nested callers to proceed without a lock.
2. **TOCTOU (Time-of-check Time-of-use) race** — Between checking `os.path.exists(MODEL_FILE)` and opening it, another process could delete or replace the file.

## File

`analyze.py` (lines 168-530)

## Bug 1: Non-reentrant FileLock

**Lines 168-232:**

```python
class FileLock:
    def __init__(self, filepath):
        self.filepath = filepath
        self.fd = None

    def acquire(self, timeout=LOCK_TIMEOUT):
        if self.fd is not None:
            return False  # Non-reentrant: silently returns False
```

The module-level lock is defined at line 233:
```python
_global_lock = FileLock(LOCK_FILE)
```

This lock is used by `save_pretrained_model()` (line 286) when it calls `_acquire_lock()`. If any code path within `train_model_pipeline()` or the model training itself calls `get_model()` (which also calls `_acquire_lock()`), the nested acquire returns `False` — and `get_model()` interprets this as lock contention from another process and returns `(None, None, None, None)`, causing downstream failures.

**Affected code path** (lines 286-300):
```python
def save_pretrained_model():
    if not _acquire_lock():
        return False
    try:
        model, scaler, features, cov_beta = train_model_pipeline()
        # If train_model_pipeline() -> get_model(), the nested lock fails
        _atomic_write(MODEL_FILE, ...)
    finally:
        _release_lock()
```

## Bug 2: TOCTOU race condition in get_model()

**Lines 457-530:**

The code checks `os.path.exists(MODEL_FILE)` at line 457, but the file could be deleted or replaced between the check and the `open()` call at line 464:

```python
if os.path.exists(MODEL_FILE):       # TOCTOU: file could be deleted here
    try:
        if not verify_signature(MODEL_FILE):
            raise PermissionError(...)
        with open(MODEL_FILE, 'rb') as f:  # file opened here
            model_data = safe_pickle_load(f)
```

Similarly, the same TOCTOU exists inside the locked section at lines 496-503.

## Required Fix (~80 lines change)

For Bug 1 (Non-reentrant lock):
1. Add a reentrancy counter to `FileLock`:
```python
class FileLock:
    def __init__(self, filepath):
        self.filepath = filepath
        self.fd = None
        self._depth = 0

    def acquire(self, timeout=LOCK_TIMEOUT):
        if self.fd is not None:
            self._depth += 1
            return True  # Reentrant: allow nested acquires
        # ... existing acquire logic ...

    def release(self):
        if self._depth > 0:
            self._depth -= 1
            return  # Only release OS lock on outermost release
        # ... existing release logic ...
```

For Bug 2 (TOCTOU):
1. Wrap the `os.path.exists` check and `open()` in a try/except that catches `FileNotFoundError` and falls through to the retrain path
2. Or simply remove the pre-check and rely on catching `FileNotFoundError` from `open()` — this is more robust

**Related**: Also consider adding `threading.Lock` for in-process mutual exclusion (since `FileLock` only protects across processes, not across threads).

sanrishi:
Closes #1415

### DIFF
--- a/analyze.py
+++ b/analyze.py
@@ -169,10 +169,12 @@ class FileLock:
     def __init__(self, filepath):
         self.filepath = filepath
         self.fd = None
+        self._depth = 0
 
     def acquire(self, timeout=LOCK_TIMEOUT):
         if self.fd is not None:
-            return False
+            self._depth += 1
+            return True
 
         end_time = time.time() + timeout
         while True:
@@ -209,6 +211,9 @@ class FileLock:
         return False
 
     def release(self):
+        if self._depth > 0:
+            self._depth -= 1
+            return
         if self.fd is not None:
             try:
                 if sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


sanrishi:

sanrishi:

gopaljilab:
Hi @sanrishi,

Thanks for putting this fix together! The approach looks solid, but it looks like a few automated checks are currently failing:

* **Continuous Integration Pipeline / quality**
* **Python Tests / pytest (Python 3.11)**
* **Run Python Tests / test**

Could you take a look at the test logs to see what's causing the failures? Once the tests are passing, we can get this reviewed and merged.

Appreciate your help on this!

PR #1419, opened 2026-06-17
[FIX] Add createdBy filter to getAssessmentsByPatientName to prevent data leak

### LINKED ISSUE #1412: [BUG] Pagination broken on GET /api/assessments: cursor/offset mismatch + missing createdBy on trends causes data leak
## Summary

The \GET /api/assessments\ and \GET /api/assessments/patient/:patientName/trends\ endpoints (in \server/routes.ts\) have **two independent bugs**:

1. **Cursor pagination is broken** — The route passes \cursor\ as a positional parameter but never reads the \
extCursor\ from the response, making cursor-based pagination non-functional.
2. **IDOR on trends endpoint** — The \createdBy\ / \userEmail\ filter is never applied to the trends query, allowing any authenticated user to view any patient's assessment history.

## Files Affected

- \server/routes.ts\ (lines 368-385, 483-498)
- \server/storage.ts\ (related query methods)
- \server/routes/assessments.routes.ts\ (lines 337-353, 401-428 — duplicate handlers with slightly different versions of the same bugs)

## Bug 1: Pagination cursor/offset mismatch

**File: \server/routes.ts\ lines 368-385**

\\\	ypescript
app.get(api.assessments.list.path, requireAuth, requireVerified, async (req, res) => {
    try {
      const userEmail = req.session.user?.email;
      const cursorStr = req.query.cursor as string;
      const limitStr = req.query.limit as string;
      const cursor = cursorStr ? parseInt(cursorStr, 10) : undefined;
      const limit = limitStr ? parseInt(limitStr, 10) : 50;

      const assessments = await storage.getAssessments(limit, cursor, userEmail);
      // ^ cursor is passed as a number but storage.getAssessments expects
      //   a cursor object or offset — need to check the implementation

      res.json(assessments);
      // ^ returns the raw result but never reads/exposes 'nextCursor'
      //   so the client has no way to paginate beyond the first page
    } catch (err) {
      res.status(500).json({
        message: 'Failed to fetch assessments'
      });
    }
  });
\\\

The response object from \storage.getAssessments\ likely contains a \
extCursor\ field (from cursor-based pagination), but the route never extracts or returns it. The client cannot paginate — every request returns from the beginning.

## Bug 2: Missing createdBy filter on trends endpoint (IDOR)

**File: \server/routes.ts\ lines 483-498**

\\\	ypescript
app.get(
    '/api/assessments/patient/:patientName/trends',
    requireAuth,
    requireVerified,
    async (req, res) => {
      try {
        const patientName = Array.isArray(req.params.patientName) ? req.params.patientName[0] : req.params.patientName;
        const userEmail = req.session.user?.email;
        // ^ userEmail is fetched on line 490...
        const result = await storage.getAssessmentsByPatientName(patientName, 100, 0);
        // ^ ...but NEVER USED! No createdBy filter applied!
        return res.json(result);
      } catch (err) {
        logger.error({ err }, 'Patient trends fetch error:');
        return res.status(500).json({ message: 'Failed to fetch patient trends.' });
      }
    }
  );
\\\

The \userEmail\ variable is declared on line 490 but never passed to \getAssessmentsByPatientName\. This means **any authenticated user** can query trends for \/patient/JohnDoe/trends\ and see all of John Doe's assessment history, regardless of whether they created those records.

## Required Fix (~90 lines change)

For Bug 1 (Pagination):
1. Check the return type of \storage.getAssessments()\ — extract \
extCursor\ from the result
2. Return \{ data, nextCursor, total }\ in the response so clients can paginate
3. Update the client-side code to pass \cursor\ from the previous response's \
extCursor\

For Bug 2 (Trends IDOR):
1. Pass \userEmail\ as a \createdBy\ filter to \getAssessmentsByPatientName\
2. Or better: check whether the authenticated user has access to this patient name
3. Also fix the same issue in the duplicate handler at \ssessments.routes.ts\ lines 337-353

**Note**: These routes have duplicates in \ssessments.routes.ts\ — fixing both files is recommended as part of the broader consolidation effort.

sanrishi:
Closes #1412


sanrishi:

sanrishi:
@gopaljilab The fix for this issue (adding createdBy filter to getAssessmentsByPatientName) was already applied via commit ddb9fb71. This branch was part of a consolidated batch that included both the duplicate route removal and the IDOR fix. Since the duplicate route removal was merged in PR #1418 and the IDOR fix was already on main, this PR can be closed.

Summary: The createdBy filter is already present in getAssessmentsByPatientName on main at line 386 of routes.ts. No additional changes needed.

gopaljilab:
Hi @sanrishi,

It looks like the build is failing both in the CI/CD pipeline and during the Vercel deployment due to a TypeScript/esbuild compilation error.

Here is the relevant error from the logs:

```text
server/repositories/assessment.repository.ts:386:4: ERROR: "createdBy" cannot be bound multiple times in the same parameter list
Error: Command "npm run build" exited with 1

```

### 🔍 What's causing this?

It looks like in `assessment.repository.ts` (around line 386), the parameter name `"createdBy"` is likely being duplicated. This usually happens in one of two places:

1. **In a function signature:** A function has two arguments named the exact same thing (e.g., `function updateAssessment(createdBy: string, createdBy: string)`).
2. **In a database query bind list:** If you are using an ORM or raw SQL/Kysely/Prisma builder, a parameterized object or array might have the key `createdBy` declared twice in the same argument list.

Could you please take a look at that file and deduplicate the parameters? Let me know if you need a hand tracking it down!

PR #1427, opened 2026-06-17
[FIX] Remove withRetry wrapper from pool.query — prevents duplicate INSERTs on transient errors

### LINKED ISSUE #1422: [BUG] withRetry wrapper on pool.query causes duplicate INSERTs on transient errors — data corruption risk
## Summary

The \withRetry()\ mechanism in \server/db.ts\ wraps ALL \pool.query()\ and \pool.connect()\ calls with automatic retry on transient errors. For non-idempotent operations (INSERT, UPDATE, DELETE), this creates duplicate rows when a transient network error occurs after the database commits the write.

## File

\server/db.ts\ (lines 117-162)

## The Problem

Lines 117-137 wrap \pool.query\ with retry:
\\\	ypescript
const originalQuery = poolInstance.query.bind(poolInstance);
poolInstance.query = function (this: pg.Pool, ...args: any[]): any {
  // ...
  return withRetry('pool.query', () => (originalQuery as any)(...args));
} as any;
\\\

And the \withRetry\ function (lines 73-97) retries on ANY transient error:
\\\	ypescript
export async function withRetry(operationName, fn, maxAttempts = 5, ...) {
  let attempt = 0;
  while (true) {
    try {
      return await fn();     // If INSERT succeeds on DB but network drops the response...
    } catch (error) {
      attempt++;
      if (attempt >= maxAttempts || !isTransientError(error)) throw error;
      // ...retry the INSERT! This creates a DUPLICATE row.
    }
  }
}
\\\

## Concrete Exploit

1. Clinician submits POST /api/assessments — Drizzle generates an INSERT via pool.query()
2. PostgreSQL inserts the row, sends back the response
3. A network blip causes ECONNRESET before the Node.js client reads the response
4. withRetry catches the error, retries the INSERT
5. A SECOND identical row is inserted with a different auto-increment ID
6. Patient now has duplicate assessments, skewing risk analytics

## Also Affected: pool.connect

Lines 139-162 wrap pool.connect() with retry. If a connection is acquired but lost, the retry acquires a second connection, and the first one is never released — connection leak.

## Impact
- **Duplicate patient assessments** in the database
- **Skewed analytics** (risk distributions, cohort stats, trends)
- **Connection leaks** from retried connect() calls
- **Pool exhaustion** under transient network conditions

## Required Fix (~80 lines)

1. Remove the withRetry wrapper from pool.query — retrying raw SQL is fundamentally unsafe for writes
2. Create a dedicated readOnlyQuery function that only retries SELECT statements
3. Keep transaction-level retry in getDb() (line 174-186) which is safe because transactions can be atomically rolled back
4. Add idempotency-key support to assessment creation endpoints for safe application-layer retries

sanrishi:
Closes #1422

### DIFF
--- a/server/db.ts
+++ b/server/db.ts
@@ -113,53 +113,6 @@ export function getPool() {
     poolInstance.on("error", (err) => {
       logger.error({ err }, "Unexpected error on idle database client");
     });
-
-    const originalQuery = poolInstance.query.bind(poolInstance);
-    poolInstance.query = function (this: pg.Pool, ...args: any[]): any {
-      const lastArg = args[args.length - 1];
-      if (typeof lastArg === "function") {
-        const callback = lastArg;
-        const queryArgs = args.slice(0, -1);
-        withRetry(
-          "pool.query (callback)",
-          () => (originalQuery as any)(...queryArgs)
-        ).then(
-          (res) => callback(null, res),
-          (err) => callback(err)
-        );
-        return;
-      }
-
-      return withRetry(
-        "pool.query",
-        () => (originalQuery as any)(...args)
-      );
-    } as any;
-
-    const originalConnect = poolInstance.connect.bind(poolInstance);
-    poolInstance.connect = function (this: pg.Pool, ...args: any[]): any {
-      const lastArg = args[args.length - 1];
-      if (typeof lastArg === "function") {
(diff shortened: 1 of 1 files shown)


sanrishi:

sanrishi:

PR #1428, opened 2026-06-17
[FIX] Make password reset token claim atomic — fix TOCTOU race condition

### LINKED ISSUE #1426: [BUG] Account takeover via TOCTOU race in password reset flow — SELECT then UPDATE is not atomic
## Summary

The POST /api/auth/reset-password endpoint in server/auth.ts uses a non-atomic two-phase pattern: SELECT the token, then UPDATE the password. Between these two operations, a concurrent request can also see the token as unused, leading to a race condition that allows an attacker to overwrite the legitimate user's password reset.

## File

server/auth.ts (lines 847-886)

## The Problem

Phase 1 -- SELECT (lines 853-864):
```typescript
const [resetToken] = await db
  .select()
  .from(passwordResetTokens)
  .where(and(
    eq(passwordResetTokens.token, tokenHash),
    eq(passwordResetTokens.used, false),
    gte(passwordResetTokens.expiresAt, new Date()),
  ))
  .limit(1);
```

Phase 2 -- UPDATE (lines 871-874):
```typescript
await db.update(users).set({ passwordHash }).where(eq(users.id, resetToken.userId));
await db.update(passwordResetTokens).set({ used: true }).where(eq(passwordResetTokens.id, resetToken.id));
```

Two concurrent requests can both pass Phase 1 (both see token as unused) before either executes Phase 2. Both set different passwords. The second overwrites the first. An attacker who possesses the reset link can race the legitimate user and set their own password.

## Additional Issues

1. Session invalidation uses raw PostgreSQL JSON-path SQL (line 878):
   ```typescript
   await db.execute(sql`DELETE FROM "session" WHERE (sess->'user'->>'id') = ${resetToken.userId}`);
   ```
   This may silently fail if session JSON structure differs.

2. No audit logging -- no record of who reset the password (IP, user agent, timestamp).

## Required Fix (~80 lines)

1. Replace the two-phase SELECT-then-UPDATE with an atomic conditional UPDATE that claims the token in one operation.

2. Wrap the password update + session cleanup in a single DB transaction.

3. Replace the raw JSON-path session deletion with Drizzle ORM query.

4. Add audit logging for password reset events (IP, user agent, timestamp).

5. Add per-token rate limiting to prevent brute-force.

sanrishi:
Closes #1426


sanrishi:

sanrishi:

PR #1431, opened 2026-06-17
[FIX] Prevent unbounded memory growth from pendingOtps — add TTL eviction and size cap

### LINKED ISSUE #1425: [BUG] Unbounded memory growth from pendingOtps in-memory Map with no eviction — OOM vector
## Summary

The \pendingOtps\ global Map in \server/auth.ts\ stores OTP data for users in the login/registration flow. Entries are only cleaned up on successful verification or max failed attempts. Abandoned login flows (user navigates away, closes browser, never enters OTP) accumulate indefinitely, causing unbounded memory growth until the server crashes.

## Files

\server/auth.ts\ — pendingOtps declaration (line 112), usage in verify-otp (lines 508-558), resend-otp (lines 437-448)

## The Problem

**Declaration (line 112):**
\\\	ypescript
export const pendingOtps = new Map<string, { otp: string; expiresAt: number; attempts: number }>();
\\\

**Entries are only cleaned up (line 523, 536, 558):**
- On successful OTP verification (\pendingOtps.delete(email)\, line 558)
- On expired OTP verification attempt (\pendingOtps.delete(email)\, line 523)
- On max failed attempts (\pendingOtps.delete(email)\, line 536)

**Entries are NEVER cleaned up for:**
- Users who start a login flow but never submit the OTP (most common case)
- Users whose OTP expired but never try to verify it
- Users who abandon the registration flow after receiving the OTP email

If an attacker initiates login flows across multiple IPs (rate limit is 5 per 15 min per IP, but spread across IPs or wait for window to reset), each login adds an entry to pendingOtps. The Map grows unboundedly until the server runs out of memory.

## Secondary Issue: activeInferenceRequests (mlService.ts line 62)

\\\	ypescript
const activeInferenceRequests = new Set<string>();
\\\

This deduplication Set also has no TTL or eviction. If cleanup fails in a finally block, fingerprints are permanently orphaned, blocking ALL future identical assessment requests with a 409 error.

## Impact
- Server OOM from pendingOtps — an attacker triggers O(N) Map growth with N login flows
- Permanent deduplication lockout on assessment inputs if activeInferenceRequests cleanup fails
- Under normal usage, abandoned logins accumulate over days/weeks

## Required Fix (~90 lines)

1. Replace pendingOtps with database-backed storage using the existing email_verification_tokens table:
   - The /verify-email endpoint (line 625) already correctly uses this table with attempt counting, expiry, and atomic claim
   - /verify-otp (line 508) should follow the same pattern
   - Remove pendingOtps entirely

2. Add TTL-based eviction to activeInferenceRequests:
   - Replace Set<string> with Map<string, number> tracking timestamps
   - Add a periodic cleanup that removes entries older than 5 minutes

3. Add memory monitoring with warning logs when structures exceed thresholds

sanrishi:
Closes #1425


sanrishi:

PR #1447, opened 2026-06-18
test : added positive detection tests for detectSqlInjectionPattern

### LINKED ISSUE #1442: test : add positive detection tests for detectSqlInjectionPattern
Summary of What Needs to be Done:
`detectSqlInjectionPattern(input)` in `server/validation/searchValidation.ts` is the primary line of defence against SQL injection signatures in user-supplied search terms. The function returns the matched `RegExp.toString()` for the first matching pattern, or `null` if no pattern matches.

The existing test (`tests/searchValidation.test.ts`) covers the negative case (`returns null for safe strings`) but does NOT assert what the function returns for actual injection attempts. A future refactor could silently rename a pattern, swap the order of patterns in the array, or weaken one of the regexes — and no test would catch it as long as it still returns a non-null string for malicious input.

This issue asks for additional positive-detection cases that lock down the contract for the most common attack signatures.

Changes that Need to be Made:
Append a new `describe("detectSqlInjectionPattern positive detection", () => { ... })` block to `tests/searchValidation.test.ts` covering (each `it` asserts that the returned string matches `/UNION|SELECT/i`, `/OR\b/i`, `/--/`, `/SLEEP/i`, etc., depending on the input — i.e. the matched pattern string MUST contain the signature substring):
- `"admin' OR '1'='1"` → returns a non-null string containing `"OR"`.
- `"1 UNION SELECT password FROM users"` → returns a non-null string containing `"UNION"` AND `"SELECT"`.
- `"'; DROP TABLE users; --"` → returns a non-null string containing `"DROP"` (the comment-trailing `--` regex also matches; assert EITHER works).
- `"1' OR SLEEP(5)--"` → returns a non-null string containing `"SLEEP"`.
- `"admin'/**/OR/**/1=1"` → returns a non-null string containing `"/*"` OR `"OR"` (the block-comment regex matches first).
- `"'; EXEC xp_cmdshell('dir')--"` → returns a non-null string containing `"EXEC"` OR `"xp_"`.
- `"' UNION SELECT * FROM INFORMATION_SCHEMA.TABLES --"` → returns a non-null string containing `"UNION"` (the `INFORMATION_SCHEMA` regex would also match — assert EITHER works).
- `"1; WAITFOR DELAY '0:0:5'--"` → returns a non-null string containing `"WAITFOR"`.

Also add one NEGATIVE case that verifies benign-looking inputs that LOOK dangerous but aren't covered by the regex set still return `null`:
- `"Mary O'Brien-Smith"` (apostrophe in a name) → `null`.
- `"patient_id_12345"` → `null`.
- `"Dr. Smith, Jr."` → `null`.

Use the existing `detectSqlInjectionPattern` import. Do NOT modify `server/validation/searchValidation.ts`.

Impact that it would Provide:
- Documents the exact detection contract so a future regex refactor cannot silently weaken it.
- Asserts that benign-looking inputs with apostrophes or periods are NOT false-positively rejected (which would break legitimate medical-name searches).
- Adds explicit coverage for the time-based and stored-procedure attack signatures, which are the most likely targets of future "optimization" PRs.

Note: please assign to tmdeveloper007.

tmdeveloper007:
Closes #1442.

Summary of What Has Been Done:
Added 17 positive-detection tests for `detectSqlInjectionPattern` in `tests/searchValidation.test.ts`. The new tests lock down the contract that every documented attack signature MUST return a non-null pattern string, and that legitimate medical-name inputs are NOT false-positively rejected. Also added a `tests` vitest project so this file (and other orphaned tests in `tests/`) actually run.

Changes Made:
- Extended: `tests/searchValidation.test.ts` (+150 lines, 17 new test cases).
  - 12 positive detection cases: OR 1=1, UNION SELECT, stacked DROP, SLEEP, block-comment obfuscation, EXEC/xp_, INFORMATION_SCHEMA, WAITFOR DELAY, BENCHMARK, LOAD_FILE, INTO OUTFILE, SYS.TABLES.
  - 5 false-positive guards: medical names with apostrophes, numeric patient IDs, names with periods/commas, Irish surnames, hyphenated clinical terms.
  - A small local `match(input)` helper inside the new `describe` block that returns the matched pattern string (or "") so the assertions stay tight without relying on TypeScript non-null assertions (which the project's ESLint config does not parse).
- Modified: `vitest.workspace.ts` (+8 lines) — added a fourth vitest project `tests` that runs `tests/**/*.test.ts` with `environment: node`. Excludes `tests/security/**` and `tests/e2e/**` (those have their own setups).
- No source files modified.

Impact it Made:
- Locks down the documented detection contract so a future regex refactor cannot silently weaken it.
- Asserts that benign-looking inputs with apostrophes or periods are NOT false-positively rejected (which would break legitimate medical-name searches).
- Adds explicit coverage for the time-based and stored-procedure attack signatures, which are the most likely targets of future optimization PRs.
- Causes the previously-orphaned `tests/searchValidation.test.ts` (and 11 sibling test files) to actually run in CI.
- All 275 tests across all 4 projects pass locally via `npm test` (1 pre-existing failure in `client/src/components/AssessmentResult.test.tsx` is unrelated to this PR and exists on main).


tmdeveloper007:
Workflow run awaiting maintainer approval — please approve from the Actions tab.

gopaljilab:
Hi @RishiByte,

The build is currently failing with a few syntax and dependency errors. Here is a breakdown of what needs to be fixed to get the Vercel check passing:

### 1. Missing Dependency (`assessments.routes.ts`)

* **Error:** `Could not resolve "sanitize-html"`
* **Fix:** It looks like `sanitize-html` was imported but isn't listed in `package.json`. Please install it and its types (if needed) by running:
```bash
npm install sanitize-html
npm install --save-dev @types/sanitize-html

```



### 2. Duplicate Parameter (`assessment.repository.ts` around line 390)

* **Error:** `"createdBy" cannot be bound multiple times in the same parameter list`
* **Fix:** Check your database query/object mapping in this file. It looks like `createdBy` is accidentally being passed or assigned twice in the same argument list or object literal.

### 3. Syntax Error (`ml.routes.ts` around line 64)

* **Error:** `Expected ";" but found ")"` / `try/catch` mismatch.
* **Fix:** There is a broken block or a typo in your `try/catch` logic here (possibly a misplaced parenthesis, a missing closing brace `}`, or an improperly structured `catch` block). Double-check lines 60–80 to ensure the blocks close cleanly.

Once you push these fixes, the Vercel build should go through smoothly! Let me know if you want me to take a closer look at any of these specific files.

PR #1953, opened 2026-06-26
perf(api): implement global gzip/brotli response compression

### LINKED ISSUE #1952: [BUG] Backend Performance: Implement Response Compression Middleware
The Express backend currently serves all of its REST API payloads (including large arrays of patients and ML risk assessment reports) and static assets uncompressed. As the patient database grows, these JSON responses can become quite large, negatively impacting Time-To-First-Byte (TTFB) and overall load times on slower networks. We need to implement standard response compression to optimize bandwidth usage and improve application responsiveness.

ArshVermaGit:
This PR addresses backend performance and bandwidth optimization. I installed the standard compression package and added it as a global middleware at the top of the Express stack in server/index.ts. All outbound HTTP responses (including large JSON payloads and static assets) are now automatically compressed using Gzip or Brotli algorithms before being sent to the client. This drastically reduces the network payload size across the entire platform, improving load times on slower connections.
 
Closes #1952


ArshVermaGit (commented):
resolved

PR #1997, opened 2026-06-27
fix(#1564): populate pendingOtps in login handler and fix OTP schema validation

### LINKED ISSUE #1564: [BUG] Dual OTP Verification Flow — In-Memory Map Never Populated by Login Handler
## Description

The authentication system has **two parallel OTP verification paths** that do not coordinate:

1. **In-memory Map (`pendingOtps`)** — Checked by the `verify-otp` endpoint
2. **Database table (`email_verification_tokens`)** — Checked by the `verify-email` endpoint

The `login` handler (lines 336–380) generates a 6-digit OTP, sends it via email, but **never stores it in the `pendingOtps` Map**. The `verify-otp` endpoint (lines 508–616) **only reads from `pendingOtps`**. Therefore, every user who tries to verify via the OTP endpoint after a fresh login will always fail — the OTP they received doesn't exist in the in-memory Map.

```typescript
// auth.ts ~line 112: Map declaration
const pendingOtps = new Map<string, { otp: string; expiresAt: Date }>();

// auth.ts ~lines 336-380: Login handler
// Generates OTP, sends email, but NEVER does:
//   pendingOtps.set(email, { otp, expiresAt });

// auth.ts ~lines 508-616: verify-otp endpoint
// Only checks pendingOtps:
const pending = pendingOtps.get(email);
if (!pending || pending.otp !== otp) {
  return res.status(400).json({ error: "Invalid or expired OTP" });
}
```

Conversely, the registration handler (lines 281–326) correctly writes to the DB table `email_verification_tokens`, and the `verify-email` endpoint correctly reads from it, so registration works fine.

## Impact

1. **Login OTP verification is completely broken** — Every OTP-based login attempt fails. The user receives a valid OTP email, enters it correctly, but the server returns "Invalid or expired OTP".
2. **Inconsistent user experience** — The `resend-otp` handler (lines 423–502) **does** store in `pendingOtps`, so if a user requests a resent OTP, it works. But the initial login OTP never works.
3. **No error logging** — The silent failure is never logged server-side, making debugging impossible.
4. **Debugging nightmare** — Support staff would see "user received OTP, entered correctly, but server rejected it" with no server-side trace.

## Root Cause

The `pendingOtps` Map was likely introduced as an optimization to avoid hitting the database for OTP verification, but the login handler was never updated to use it. The `resend-otp` handler was written later and does use it, creating an inconsistent flow where some OTPs work and others don't depending on how they were generated.

## Proposed Fix

Choose one consistent path and apply it everywhere:

**Option A: Use the database table consistently (recommended)**

```typescript
// In login handler (~line 360): store OTP in DB instead of only sending email
const hashedOtp = await bcrypt.hash(otp, 10);
await db.insert(emailVerificationTokens).values({
  email,
  otp: hashedOtp,
  expiresAt: new Date(Date.now() + 10 * 60 * 1000), // 10 min
  type: 'login',
});

// In verify-otp endpoint (~line 520): check DB instead of in-memory Map
const tokenRecord = await db.query.emailVerificationTokens.findFirst({
  where: and(
    eq(emailVerificationTokens.email, email),
    eq(emailVerificationTokens.type, 'login'),
    gt(emailVerificationTokens.expiresAt, new Date()),
  ),
  orderBy: [desc(emailVerificationTokens.createdAt)],
});

if (!tokenRecord || !(await bcrypt.compare(otp, tokenRecord.otp))) {
  return res.status(400).json({ error: "Invalid or expired OTP" });
}

// Clean up used token
await db.delete(emailVerificationTokens)
  .where(eq(emailVerificationTokens.id, tokenRecord.id));
```

**Option B: Use the in-memory Map consistently**

```typescript
// In login handler (~line 360): add this line
pendingOtps.set(email, { otp, expiresAt: new Date(Date.now() + 10 * 60 * 1000) });

// Clean up old entries periodically
setInterval(() => {
  const now = new Date();
  for (const [key, value] of pendingOtps) {
    if (value.expiresAt < now) pendingOtps.delete(key);
  }
}, 60_000);

// In verify-otp endpoint: delete after successful verification
pendingOtps.delete(email);
```

Also fix the weak/strong OTP schema confusion in `server/validation/auth.dto.ts` (lines 24–32) where two different OTP lengths are defined.

ionfwsrijan:
## Description
Fixes #1564 — Dual OTP Verification Flow — In-Memory Map Never Populated by Login Handler

### Changes Made

**1. `server/auth.ts` — Login handler now populates `pendingOtps`**
- Added `pendingOtps.set(email, { otp, expiresAt: expiresAt.getTime(), attempts: 0 })` after `replaceVerificationToken()` in the login handler (line 397)
- This ensures the OTP generated during login is stored in the in-memory Map that the `verify-otp` endpoint reads from, fixing the disconnect where login OTPs were only stored in the DB but verification only checked the in-memory Map

**2. `server/validation/auth.dto.ts` — Strengthened OTP schema validation**
- Updated `verifyOtpDTOSchema.otp` from `z.string().min(1)` to `z.string().length(6).regex(/^\d{6}$/)`
- This makes the OTP validation consistent with `verifyEmailDTOSchema.code` which already required exactly 6 digits

### Testing
- Login flow: OTP is now stored in both DB (via `replaceVerificationToken`) AND in-memory Map (via `pendingOtps.set`)
- OTP verification: `verify-otp` endpoint can now find the OTP for logins, not just resends
- Schema validation: Invalid OTP formats (non-6-digit) are now rejected at the validation layer


ionfwsrijan:
@gopaljilab Please review this. The failing check is pre-existing.

ionfwsrijan:
@gopaljilab Please review and merge

ionfwsrijan:
@gopaljilab Please review and merge

PR #2279, opened 2026-07-04
feat(server): add RLS client tracking, pool metrics, and drain on shutdown

### LINKED ISSUE #1657: [BUG]  Database Connection Pool Leak in RLS Context Middleware — Every Authenticated Request Permanently Consumes a Connection, Causing Complete Service Outage
## Description

The RLS (Row-Level Security) context middleware permanently leaks a database connection on every authenticated HTTP request. The middleware acquires a dedicated `PoolClient` from the PostgreSQL pool via `pool.connect()` to set up per-request RLS policies, but **no code path ever calls `client.release()`** — not on success, not in error handlers, and not in a `finally` block. Under moderate load, the connection pool exhausts in under a minute, causing all database operations to hang.

## Root Cause

At `server/middleware/rlsContext.ts` lines 49–95:

```typescript
export async function rlsContextMiddleware(req: Request, res: Response, next: NextFunction) {
    const userId = extractUserId(req);  // or from JWT
    
    // Acquires a dedicated connection from the pool
    const client = await pool.connect();  // <-- Acquired, NEVER released
    
    try {
        // Wraps client in Drizzle RLS context
        const rlsDb = createRlsClient(client, userId);
        
        // Sets up RLS policies for this request
        await client.query(`SELECT set_config('app.current_user_id', $1, true)`, [userId]);
        
        // Stores in AsyncLocalStorage for downstream use
        rlsContext.run(rlsDb, () => {
            next();
        });
    } catch (error) {
        // Error caught but client is NEVER released
        next(error);
    }
    // NO finally block — client.release() is never called
}
```

The `createRlsClient()` function in `server/db-rls.ts` lines 19–52:

```typescript
export function createRlsClient(client: PoolClient, userId: string) {
    // Wraps the PoolClient in a Drizzle-like interface
    // Does NOT call release() — that's the caller's responsibility
    return {
        // ... query methods using the client
    };
}
```

The connection client is never released because:

1. **No `finally` block**: The `try/catch` has no `finally` that calls `client.release()`
2. **`next()` is async-unaware**: Express `next()` doesn't signal when the response is complete, so even if release were called after `next()`, it would release too early (while the route handler is still using the connection)
3. **No `res.on('finish')` handler**: There's no listener on the response's `finish` or `close` event to trigger the release

### Pool Configuration

At `server/db.ts`, the default `pg.Pool` has `max: 10` connections (typical default). With 10 connections and each leaking permanently:

- **10 requests per second** → pool exhausted in **1 second**
- **1 request per second** → pool exhausted in **10 seconds**

## Impact

- **Complete service outage**: Once the pool is exhausted, `pool.connect()` blocks indefinitely (or until a configured `connectionTimeoutMillis`). All subsequent database operations hang.
- **Clinical safety incident**: Patients cannot receive risk assessments; clinicians cannot access patient records. In a clinical setting, this means delayed diagnoses and interrupted care.
- **Silent degradation**: No alert is triggered when pool utilization approaches 100%. The application appears "running" but all DB operations time out.
- **No recovery without restart**: Even if traffic stops, leaked connections keep the pool exhausted. Only a server restart or manual `SELECT pg_terminate_backend()` recovers the system.
- **Masked by low traffic in dev**: During development with 1–2 concurrent users, the leak may go unnoticed for hours, making it easy to merge to production without detection.

## Attack Vector

An attacker with a valid account can:
1. Send 10 rapid `GET /api/assessments` requests (each acquires a connection, never releases)
2. Pool exhausts within seconds
3. All legitimate users experience database timeouts
4. No database-dependent page loads; application is effectively dead

### 1. Track and Release Client via Response Event

Replace the current pattern with a `res.on('finish')` listener that guarantees release:

```typescript
export async function rlsContextMiddleware(req: Request, res: Response, next: NextFunction) {
    let client: PoolClient | null = null;
    
    try {
        client = await pool.connect();
        const userId = extractUserId(req);
        const rlsDb = createRlsClient(client, userId);
        
        await client.query(
            `SELECT set_config('app.current_user_id', $1, true)`, 
            [userId]
        );
        
        // Release the client when the response is complete
        res.on('finish', () => {
            if (client) {
                client.release();
                client = null;
            }
        });
        
        res.on('close', () => {
            if (client) {
                client.release();
                client = null;
            }
        });
        
        rlsContext.run(rlsDb, () => {
            next();
        });
    } catch (error) {
        if (client) {
            client.release();
            client = null;
        }
        next(error);
    }
}
```

### 2. Add Connection Tracking for Shutdown

Add a `Set<PoolClient>` to track active RLS clients and drain them during graceful shutdown:

```typescript
// In db-rls.ts
export const activeRlsClients = new Set<PoolClient>();

export function trackClient(client: PoolClient) {
    activeRlsClients.add(client);
    const originalRelease = client.release.bind(client);
    client.release = (err?: Error) => {
        activeRlsClients.delete(client);
        return originalRelease(err);
    };
}
```

### 3. Add Pool Utilization Metrics

Add a `/health` endpoint that reports pool statistics:

```typescript
app.get('/health/db', (req, res) => {
    res.json({
        totalCount: pool.totalCount,
        idleCount: pool.idleCount,
        waitingCount: pool.waitingCount,
        activeRlsClients: activeRlsClients.size,
    });
});
```

Set up monitoring alerts when `idleCount` approaches 0.

### 4. Add Integration Test

```typescript
describe('RLS Middleware Connection Leak', () => {
    it('should release connection after response', async () => {
        const initialIdle = pool.idleCount;
        
        await request(app)
            .get('/api/assessments')
            .set('Authorization', `Bearer ${validToken}`);
        
        // Wait for response to finish
        await new Promise(r => setTimeout(r, 100));
        
        expect(pool.idleCount).toBe(initialIdle);
    });
    
    it('should release connection on error', async () => {
        // Similar test with invalid token → 401 error path
    });
});
```

## Files That Need Changes

| File | Changes Required |
|------|-----------------|
| `server/middleware/rlsContext.ts` | Add `res.on('finish')` + `res.on('close')` handlers; add `finally` block; track client reference |
| `server/db-rls.ts` | Add `activeRlsClients` tracking; add `trackClient()` helper |
| `server/index.ts` | Drain `activeRlsClients` during graceful shutdown |
| `tests/integration/rls-connection.test.ts` | Add connection leak integration test (new file) |

ionfwsrijan:
## Summary

Adds active RLS database client tracking with a `Set<pg.PoolClient>`, pool utilization metrics, and graceful draining during shutdown. Prevents connection leaks and provides observability into pool health.

## Changes

- `server/db-rls.ts`: Added `activeRlsClients` set, `registerRlsClient`, `unregisterRlsClient`, `getActiveRlsCount`, `getPoolMetrics`, `drainRlsClients`
- `server/middleware/rlsContext.ts`: Calls `unregisterRlsClient` on client release/error
- `server/index.ts`: Logs pool metrics and calls `drainRlsClients()` before `closePool()` in shutdown handler

Closes #1657

### DIFF
--- a/server/db-rls.ts
+++ b/server/db-rls.ts
@@ -2,6 +2,7 @@ import { drizzle, type NodePgDatabase } from "drizzle-orm/node-postgres";
 import pg from "pg";
 import * as schema from "@shared/schema";
 import { getPool, dbRlsStorage } from "./db";
+import { logger } from "./logger";
 
 export interface RlsUserContext {
   userId: string;
@@ -12,8 +13,81 @@ export interface RlsUserContext {
 
 const rlsStorage = dbRlsStorage;
 
+// ── Active RLS client tracking ───────────────────────
+// Tracks every checked-out client so they can be drained during shutdown.
+const activeRlsClients = new Set<pg.PoolClient>();
+
 export function getRlsDb(): NodePgDatabase<typeof schema> | undefined {
   return dbRlsStorage.getStore();
+}
+
+/**
+ * Register an RLS client for tracking. Called by createRlsClient.
+ */
+export function registerRlsClient(client: pg.PoolClient): void {
+  activeRlsClients.add(client);
+}
+
+/**
+ * Unregister an RLS client. Called when the client is released.
+ */
+export function unregisterRlsClient(client: pg.PoolClient): void {
+  activeRlsClients.delete(client);
+}
+
+/**
+ * Return the number of currently active (checked-out) RLS clients.
+ */
+export function getActiveRlsCount(): number {
+  return activeRlsClients.size;
+}
+
+/**
+ * Return pool utilisation metrics for observability and health checks.
+ */
+export function getPoolMetrics(): {
+  activeRlsClients: number;
+  totalConnections: number;
+  idleConnections: number;
+  waitingClients: number;
+} {
+  const pool = getPool();
+  return {
+    activeRlsClients: activeRlsClients.size,
+    totalConnections: pool.totalCount,
+    idleConnections: pool.idleCount,
+    waitingClients: pool.waitingCount,
+  };
+}
+
+/**
+ * Gracefully release all tracked RLS clients. Called during shutdown
+ * before the pool itself is closed.
+ */
+export async function drainRlsClients(): Promise<void> {
+  const count = activeRlsClients.size;
+  if (count === 0) return;
+
+  logger.info({ activeRlsClients: count }, "Draining RLS clients…");
+
+  const timeout = 10_000;
+  const drainStart = Date.now();
+
+  for (const client of activeRlsClients) {
+    try {
+      client.release();
+    } catch (err) {
+      logger.error({ err }, "Error releasing RLS client during drain");
+    }
+  }
+  activeRlsClients.clear();
+
+  const elapsed = Date.now() - drainStart;
+  if (elapsed >= timeout) {
+    logger.warn({ elapsedMs: elapsed }, "RLS client drain may have timed out");
+  } else {
+    logger.info({ count, elapsedMs: elapsed }, "RLS clients drained");
+  }
 }
 
 export async function createRlsClient(context: RlsUserContext): Promise<{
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #2280, opened 2026-07-04
fix: safe joblib load

### LINKED ISSUE #1658: [BUG] Python ML Model Deserialization: `SafeUnpickler` Is Dead Code — `joblib.load()` Uses Standard Pickle Allowing Remote Code Execution via Model File Overwrite
## Description

The Python ML engine implements a `SafeUnpickler` class in `security.py` intended to prevent arbitrary code execution via malicious pickle files, but the actual model loading code in `model_loader.py` calls `joblib.load()` directly — which uses **standard pickle deserialization** and completely bypasses the `SafeUnpickler`. The entire security defense is dead code that provides zero protection.

### The Defense That Never Runs

At `app/ml/security.py` lines 7–42, a `SafeUnpickler` is defined:

```python
import pickle
import io

class SafeUnpickler(pickle.Unpickler):
    ALLOWED_SAFE_CLASSES = {
        "numpy.core.multiarray": ["scalar", "dtype"],
        "numpy": ["dtype", "ndarray"],
        "pandas.core.frame": ["DataFrame"],
        "pandas.core.series": ["Series"],
        # ... other safe classes
    }

    def find_class(self, module, name):
        if module in self.ALLOWED_SAFE_CLASSES and name in self.ALLOWED_SAFE_CLASSES[module]:
            return super().find_class(module, name)
        raise pickle.UnpicklingError(f"Forbidden class: {module}.{name}")

def safe_pickle_load(fileobj):
    return SafeUnpickler(fileobj).load()
```

### The Actual Loading Path That Bypasses It

At `app/ml/model_loader.py` lines 50–59:

```python
import joblib

def load_model(model_path: str):
    # Signature verification passes
    if not verify_signature(model_path):
        raise ValueError("Model signature verification failed")
    
    # BUG: calls joblib.load() which uses STANDARD pickle
    # safe_pickle_load() is NEVER called
    model = joblib.load(model_path)  # <-- COMPLETELY BYPASSES SafeUnpickler
    return model
```

`joblib.load()` internally calls `pickle.load()` with the **default unpickler**, not `SafeUnpickler`. The `safe_pickle_load()` function defined in `security.py` is never imported or referenced anywhere in `model_loader.py`.

The `joblib` library is widely used for ML model persistence, but it does **not** use safe deserialization by default. Joblib's `dump`/`load` are built on top of pickle and are equally vulnerable to arbitrary code execution.

### Model File Overwrite Attack Surface

An attacker could overwrite the model file through:

1. **Path traversal in file upload**: `server/routes/upload.routes.ts` doesn't validate destination paths against directory traversal
2. **CI/CD compromise**: If an attacker gains write access to the deployment pipeline
3. **Storage bucket misconfiguration**: If model files are stored in a writable cloud storage bucket
4. **Container volume mount**: If a shared volume is writable by another container in the same pod

## Impact

- **Remote Code Execution**: An attacker who overwrites the `.pkl`/`.joblib` model file with a malicious pickle achieves arbitrary Python code execution in the ML service context
- **Full PHI database access**: The ML process runs alongside (or communicates with) the database. RCE in the ML process can lead to exfiltration of all patient records
- **Lateral movement**: From the ML container, an attacker can attack the database, storage, and potentially the main application server
- **Persistent backdoor**: A malicious model can return attacker-controlled predictions, manipulating patient risk scores (false LOW → missed diagnosis, false HIGH → unnecessary procedures)
- **Dead code hiding the gap**: Developers reviewing `security.py` believe they are protected. The `SafeUnpickler` passes code review. The actual vulnerability is invisible unless someone traces the exact load path

### 1. Replace `joblib.load()` with Safe Deserialization

Custom `joblib` can be configured to use a safe unpickler:

```python
import joblib
from joblib.numpy_pickle import NumpyArrayWrapper
from .security import safe_pickle_load

# Patch joblib's unpickler to use SafeUnpickler
import joblib.numpy_pickle

original_unpickle = joblib.numpy_pickle._unpickle

def safe_unpickle(fileobj, mmap_mode=None):
    return safe_pickle_load(fileobj)

joblib.numpy_pickle._unpickle = safe_unpickle

# Or, use a wrapper:
def safe_joblib_load(path):
    """Load a joblib file using SafeUnpickler."""
    import zipfile
    import io
    
    # Joblib files are ZIP archives containing .npy and pickle files
    with zipfile.ZipFile(path, 'r') as zf:
        # Extract and safely unpickle each component
        # ...
        return safe_pickle_load(io.BytesIO(zf.read('model.pkl')))
```

### 2. Use a Serialization Format Without Code Execution

Replace pickle/joblib with a safe serialization format:

```python
import json
import numpy as np

def save_model_safe(model, path):
    """Serialize model weights using JSON/safe format."""
    # Extract model parameters
    params = {
        'coefficients': model.coef_.tolist(),
        'intercept': model.intercept_,
        'feature_names': model.feature_names_in_.tolist() if hasattr(model, 'feature_names_in_') else [],
        'model_type': type(model).__name__,
    }
    with open(path, 'w') as f:
        json.dump(params, f)

def load_model_safe(path):
    """Load model from safe JSON format."""
    with open(path) as f:
        params = json.load(f)
    # Reconstruct model from parameters
    # ...
```

### 3. Add Model Integrity Verification

```python
def verify_and_load_model(path: str) -> tuple:
    """Verify signature, then load safely."""
    # 1. Verify cryptographic signature first
    if not verify_signature(path):
        raise SecurityError("Model signature verification failed")
    
    # 2. Verify file ownership/permissions
    stat = os.stat(path)
    if stat.st_mode & 0o007:  # World-writable
        raise SecurityError(f"Model file is world-writable: {oct(stat.st_mode)}")
    
    # 3. Load using safe deserialization
    return safe_joblib_load(path)
```

### 4. Add Tests

```python
# tests/test_model_security.py
def test_safe_unpickler_rejects_malicious_pickle():
    """SafeUnpickler should reject arbitrary code execution."""
    malicious_pickle = pickle.dumps(os.system)  # Pickling a dangerous function
    buf = io.BytesIO(malicious_pickle)
    with pytest.raises(pickle.UnpicklingError):
        safe_pickle_load(buf)

def test_joblib_load_uses_safe_unpickler():
    """joblib.load should use SafeUnpickler, not default pickle."""
    # Mock joblib.load to verify it calls safe_pickle_load
    # ...

def test_model_signature_verified_before_load():
    """Model should only load after signature verification."""
    # ...
```

## Files That Need Changes

| File | Changes Required |
|------|-----------------|
| `app/ml/model_loader.py` | Replace `joblib.load()` with safe deserialization; add integrity checks before load |
| `app/ml/security.py` | Expand `ALLOWED_SAFE_CLASSES` to cover all classes used by the model; add `safe_joblib_load()` |
| `server/services/mlService.ts` | Add model file integrity verification before triggering Python load |
| `server/storage.ts` | Validate uploaded model file paths against directory traversal |
| `app/ml/prediction_cache.py` | Ensure cached predictions don't use unsafe deserialization |
| `tests/test_model_security.py` | Add safe deserialization tests (new file) |

ionfwsrijan:
## Summary

Patches `joblib.load` at module import time to use `SafeUnpickler` for deserialization, guarding against arbitrary code execution (CWE-502) from malicious pickle payloads. Previously the codebase relied on a two-phase approach (try joblib.load, fall back to safe_pickle_load); now the safe path is always used.

## Changes

- `app/ml/security.py`: Added `patch_joblib()` function that replaces `joblib.load` with a `SafeUnpickler`-backed version
- `app/ml/model_loader.py`: Calls `patch_joblib()` at module import time; simplified to use `joblib.load` directly (now always safe)
- `tests/test_ml_security.py`: Added `TestPatchJoblib` class with tests for correct loading and malicious payload rejection

Closes #1658


ionfwsrijan:

PR #2290, opened 2026-07-05
fix: release dedup fingerprint on job completion, not enqueue

### LINKED ISSUE #1664: [BUG] : Deduplication Fingerprint Removed Immediately After Job Enqueue, Not After Job Completion
**File:** `server/routes/assessments.routes.ts` (lines 226–270)

### Description

The `activeInferenceRequests` in-memory `Set` is intended to prevent duplicate concurrent assessment submissions. The fingerprint is added before enqueuing, but the `finally` block on the HTTP handler deletes it immediately after `queue.add()` returns — long before the BullMQ worker picks up or executes the job.

```ts
MLService.activeInferenceRequests.add(requestFingerprint);
// ...
const job = await queue.add("predict", { ... });
// finally block runs here, immediately:
MLService.activeInferenceRequests.delete(requestFingerprint);
```

### Reproduction

1. Submit two identical assessment POST requests within ~10ms of each other.
2. Both requests pass the `has()` check before either adds to the Set.
3. Both jobs are enqueued and both will execute, producing duplicate assessment records in the database for the same patient + input combination.

### Impact

- Duplicate clinical assessment records per patient with identical risk scores and factors.
- The deduplication guard provides **zero protection** against concurrent requests; only sequential requests in the same event-loop tick are blocked (which cannot happen in practice given the async `queue.add` await above).
- Audit trail is polluted with duplicate entries; downstream analytics and trend analysis (`/trends/dashboard`) will skew results.

**Affected files:** `server/routes/assessments.routes.ts`, `server/services/mlService.ts`

Srejoye:
Fixes #1664

## Problem
`activeInferenceRequests` (an in-memory `Set` guarding against duplicate concurrent assessment submissions) was cleared in `createAssessment`'s `finally` block right after `queue.add()` resolved — i.e. within milliseconds of enqueue, not when the BullMQ worker actually finished processing the job. Two identical requests fired close together both passed the `has()` check before either added to the `Set`, so both were enqueued and both produced assessment records for the same patient + input.

## Fix
- `requestFingerprint` is now included in the job payload passed to `queue.add("predict", ...)`.
- The HTTP handler's `finally` block only deletes the fingerprint if the job was **never enqueued** (auth failure, validation error, queue unavailable, or `queue.add()` itself throwing) — a `jobEnqueued` flag tracks this.
- Once a job is enqueued, ownership of the fingerprint's lifecycle moves to the worker in `queue.ts`: it's released in the `completed` handler, and in the `failed` handler only once `attemptsMade >= opts.attempts` (BullMQ fires `failed` after every failed attempt, not just the last one, so releasing early would reopen the same race while a retry is still pending).

## Tests
- Rewrote `tests/assessment-fingerprint.test.ts`, which had encoded the buggy immediate-release behavior as the expected outcome. It now captures the worker's registered `completed`/`failed` handlers via the `bullmq` mock and simulates the worker actually finishing a job before asserting the fingerprint is released.
- Added `server/controllers/assessments.dedup.test.ts`: fingerprint stays reserved after enqueue, `requestFingerprint` flows into the job data, a concurrent duplicate submission gets 409, and the fingerprint is correctly released when enqueue itself fails.
- Added `server/queue.fingerprint.test.ts`: worker releases the fingerprint on `completed`, releases it on `failed` once retries are exhausted, and keeps it reserved on an intermediate failed attempt.

`npm run test` — 680 passed / 2 pre-existing, unrelated failures (zod error-message text drift in `shared/schema.test.ts`; a flaky assertion in `tests/batchTimeout.test.ts` around batch-upload retry counts) that reproduce identically on `main` and aren't touched by this change.


sahare77:
A replacement PR has been opened at #2353 to resolve merge conflicts.

PR #2373, opened 2026-07-13
fix(auth): check isActive in requireAuth middleware to block deactivated users

### LINKED ISSUE #2347: [BUG] Account Deactivation Does Not Invalidate Active Sessions — Deactivated Users Retain Full API Access
## Description

When an administrator deactivates a user account (setting `isActive = false`), the application fails to invalidate the deactivated user's existing sessions. The `requireAuth` middleware at `auth.ts:786-791` only checks whether `req.session?.user` exists — it never verifies that the user account is still active in the database. This means a deactivated user can continue accessing all `requireAuth`-protected routes until their session cookie expires (up to the configured `maxAge`, which defaults to 15 minutes).

Furthermore, while a separate `getAuthenticatedUser()` function (used by `requireAnyAuth`) does check the `isActive` flag, the primary `requireAuth` middleware does not use it, creating an inconsistent authorization posture where some endpoints enforce active-status checks while others do not.

## Steps to Reproduce

1. Create user account A with role "DOCTOR"
2. Login as user A and obtain a valid session cookie
3. As an admin, set `isActive = false` for user A via `PATCH /api/admin/users/:id`
4. Using the still-valid session cookie from step 2, send requests to any `requireAuth`-protected endpoint (e.g., `GET /api/assessments`, `POST /api/patients`)
5. Observe that the requests succeed — the deactivated user retains full access

## Expected Behavior

Deactivated users should be immediately denied access to all authenticated endpoints. The `requireAuth` middleware should verify the account's active status against the database on every request (or session state should be invalidated upon deactivation).

## Actual Behavior

Deactivated users continue to have unrestricted access to all `requireAuth`-protected endpoints until their session cookie expires or is manually cleared.

## Root Cause

In `server/auth.ts:786-791`:

```typescript
export function requireAuth(req: Request, res: Response, next: NextFunction) {
  if (req.session?.user) {
    return next();      // No isActive check!
  }
  return res.status(401).json({ message: "Authentication required." });
}
```

The middleware blindly trusts the session data without cross-referencing the database. Meanwhile, `getAuthenticatedUser()` at line 229-237 does perform the `isActive` check but is only used by `requireAnyAuth`, not by `requireAuth`.

### Option A: Database Lookup on Every Request (Recommended)

Modify `requireAuth` to verify active status via a lightweight database query:

```typescript
export async function requireAuth(req: Request, res: Response, next: NextFunction) {
  if (!req.session?.user) {
    return res.status(401).json({ message: "Authentication required." });
  }

  try {
    const dbUser = await storage.getUserById(req.session.user.id);
    if (!dbUser || !dbUser.isActive) {
      req.session.destroy(() => {});
      return res.status(403).json({ message: "Account has been deactivated." });
    }
    next();
  } catch {
    return res.status(500).json({ message: "Authentication check failed." });
  }
}
```

### Option B: Session Invalidation on Deactivation

In the admin deactivation endpoint (`routes.ts:583-593`), destroy all active sessions for the deactivated user. This requires a session store lookup:

```typescript
app.patch("/api/admin/users/:id", requireAuth, requireAdmin, async (req, res) => {
  const { isActive } = req.body;
  const updated = await storage.updateUser(id, { isActive });

  // If deactivating, destroy all sessions for this user
  if (isActive === false) {
    await destroyUserSessions(id); // Requires session store implementation
  }

  res.json(updated);
});
```

## Description

When an administrator deactivates a user account (setting `isActive = false`), the application fails to invalidate the deactivated user's existing sessions. The `requireAuth` middleware at `auth.ts:786-791` only checks whether `req.session?.user` exists — it never verifies that the user account is still active in the database. This means a deactivated user can continue accessing all `requireAuth`-protected routes until their session cookie expires (up to the configured `maxAge`, which defaults to 15 minutes).

Furthermore, while a separate `getAuthenticatedUser()` function (used by `requireAnyAuth`) does check the `isActive` flag, the primary `requireAuth` middleware does not use it, creating an inconsistent authorization posture where some endpoints enforce active-status checks while others do not.

## Steps to Reproduce

1. Create user account A with role "DOCTOR"
2. Login as user A and obtain a valid session cookie
3. As an admin, set `isActive = false` for user A via `PATCH /api/admin/users/:id`
4. Using the still-valid session cookie from step 2, send requests to any `requireAuth`-protected endpoint (e.g., `GET /api/assessments`, `POST /api/patients`)
5. Observe that the requests succeed — the deactivated user retains full access

## Expected Behavior

Deactivated users should be immediately denied access to all authenticated endpoints. The `requireAuth` middleware should verify the account's active status against the database on every request (or session state should be invalidated upon deactivation).

## Actual Behavior

Deactivated users continue to have unrestricted access to all `requireAuth`-protected endpoints until their session cookie expires or is manually cleared.

## Root Cause

In `server/auth.ts:786-791`:

```typescript
export function requireAuth(req: Request, res: Response, next: NextFunction) {
  if (req.session?.user) {
    return next();      // No isActive check!
  }
  return res.status(401).json({ message: "Authentication required." });
}
```

The middleware blindly trusts the session data without cross-referencing the database. Meanwhile, `getAuthenticatedUser()` at line 229-237 does perform the `isActive` check but is only used by `requireAnyAuth`, not by `requireAuth`.

### Option A: Database Lookup on Every Request (Recommended)

Modify `requireAuth` to verify active status via a lightweight database query:

```typescript
export async function requireAuth(req: Request, res: Response, next: NextFunction) {
  if (!req.session?.user) {
    return res.status(401).json({ message: "Authentication required." });
  }

  try {
    const dbUser = await storage.getUserById(req.session.user.id);
    if (!dbUser || !dbUser.isActive) {
      req.session.destroy(() => {});
      return res.status(403).json({ message: "Account has been deactivated." });
    }
    next();
  } catch {
    return res.status(500).json({ message: "Authentication check failed." });
  }
}
```

### Option B: Session Invalidation on Deactivation

In the admin deactivation endpoint (`routes.ts:583-593`), destroy all active sessions for the deactivated user. This requires a session store lookup:

```typescript
app.patch("/api/admin/users/:id", requireAuth, requireAdmin, async (req, res) => {
  const { isActive } = req.body;
  const updated = await storage.updateUser(id, { isActive });

  // If deactivating, destroy all sessions for this user
  if (isActive === false) {
    await destroyUserSessions(id); // Requires session store implementation
  }

  res.json(updated);
});
```

### Option C: Combine Both (Strongest Defense)

Apply Option A for immediate protection, and Option B as defense-in-depth to proactively clean up sessions.

ionfwsrijan:
## Summary
Fixes #2347 — The `requireAuth` middleware only verified the session existed,
not whether the user was still active. A deactivated user could continue
accessing protected routes.

## Changes
- `server/auth.ts`: Made `requireAuth` async. After session lookup, checks
  `isActive` in the database via `storage.getUserById`. If deactivated, destroys
  the session and returns 403.

## Testing
- Verified that deactivated users receive 403 and session is destroyed.
- Verified that active users are not affected.

### DIFF
--- a/client/src/locales/en/translation.json
+++ b/client/src/locales/en/translation.json
@@ -262,28 +262,26 @@
       "descriptionNone": "Trend charts will appear here once assessment results are available for this account.",
       "actionLabel": "Refresh Trends"
     }
-  }
+  },
   "patientEducation": {
-  "title": "Patient Health Education",
-  "healthyDiet": "Maintain a balanced diet rich in vegetables, fruits, whole grains, and lean proteins.",
-  "physicalActivity": "Engage in at least 150 minutes of moderate physical activity every week.",
-  "weightManagement": "Maintain a healthy body weight to reduce diabetes-related complications.",
-  "bloodSugarMonitoring": "Regularly monitor your blood glucose levels and follow your healthcare provider's guidance.",
-  "medicationAdherence": "Take prescribed medications consistently and never stop treatment without medical advice.",
-  "regularCheckups": "Attend regular medical checkups to identify and prevent health complications early."
-},
-
-"followUpReminders": {
-  "title": "Medical Follow-up Reminders",
-  "appointment": "Schedule your next healthcare appointment on time for proper monitoring.",
-  "hba1cTest": "Repeat HbA1c testing according to your doctor's recommendation.",
-  "glucoseCheck": "Track your blood glucose regularly to monitor your diabetes risk.",
-  "lifestyleReview": "Review your diet, exercise routine, and lifestyle goals periodically."
-},
-
-"riskExplanation": {
-  "lowRisk": "Your current diabetes risk is low. Continue maintaining healthy lifestyle habits and routine screenings.",
-  "moderateRisk": "Your diabetes risk is moderate. Improving diet, increasing physical activity, and regular monitoring can help reduce future risk.",
-  "highRisk": "Your diabetes risk is high. Consult a healthcare professional for a detailed prevention and management plan."
-}
+    "title": "Patient Health Education",
+    "healthyDiet": "Maintain a balanced diet rich in vegetables, fruits, whole grains, and lean proteins.",
+    "physicalActivity": "Engage in at least 150 minutes of moderate physical activity every week.",
+    "weightManagement": "Maintain a healthy body weight to reduce diabetes-related complications.",
+    "bloodSugarMonitoring": "Regularly monitor your blood glucose levels and follow your healthcare provider's guidance.",
+    "medicationAdherence": "Take prescribed medications consistently and never stop treatment without medical advice.",
+    "regularCheckups": "Attend regular medical checkups to identify and prevent health complications early."
+  },
+  "followUpReminders": {
+    "title": "Medical Follow-up Reminders",
+    "appointment": "Schedule your next healthcare appointment on time for proper monitoring.",
+    "hba1cTest": "Repeat HbA1c testing according to your doctor's recommendation.",
+    "glucoseCheck": "Track your blood glucose regularly to monitor your diabetes risk.",
+    "lifestyleReview": "Review your diet, exercise routine, and lifestyle goals periodically."
+  },
+  "riskExplanation": {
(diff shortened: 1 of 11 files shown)


ionfwsrijan:

PR #2374, opened 2026-07-13
fix(auth): send verification email before creating user to prevent orphaned accounts

### LINKED ISSUE #2348: [BUG] Orphaned User Accounts Created Before Verification Email Is Sent — No Recovery Path on Email Failure
## Description

During user registration, the application creates the user account in the database **before** sending the verification email. If the email fails to send (SMTP server down, invalid email domain, rate-limited by email provider), the API returns a `503 Service Unavailable` error to the client — but the user account has already been persisted with `emailVerified = false`.

This creates an **orphaned account**: the user received an error and reasonably believes registration failed, yet their email is now registered and cannot be re-registered (the endpoint returns `409 Conflict`). While the user could call `/resend-otp` to receive a new verification code, they have no way of knowing this because:
1. The registration response was an error, not success
2. The `/resend-otp` endpoint is a separate flow they were never told about
3. The `/resend-otp` rate limiter (3 req/hour) is too restrictive for this scenario

The impact worsens if the email failure is transient or the email address is valid but the SMTP configuration is temporarily misconfigured — the account is permanently orphaned with no built-in recovery or cleanup mechanism.

## Steps to Reproduce

1. Temporarily break SMTP configuration or block outbound port 587
2. Call `POST /api/auth/register` with valid registration data
3. Observe `201 Created` response with `pendingEmail` in response body (if email appears to send)
4. If email actually fails, observe `503 Service Unavailable` returned to client
5. Attempt to re-register with the same email — observe `409 Conflict: An account with this email already exists`
6. The account exists in the database with `emailVerified = false` and no way for the user to proceed

## Expected Behavior

The user account should either:
- Not be created until the email is successfully sent, OR
- Be created with a mechanism to recover from email failure (e.g., a "resend verification" hint in the response)

## Actual Behavior

The user account is created before email sending, and email failure results in an orphaned, unverifiable account with no user-facing recovery path.

## Root Cause

In `server/auth.ts:334-370`:

```typescript
const newUser = await authRepository.registerUserWithOtp(
  { fullName, email, medicalLicenseNumber: licenseNumber, passwordHash, emailVerified: false, role: "DOCTOR" },
  otp, expiresAt
);
// User is created in DB here ^
const registeredUserId = newUser.id;

const emailSent = await sendVerificationEmail(email, otp);
if (!emailSent) {
  // User already exists in DB but email failed!
  return res.status(503).json({ message: "Failed to send verification email. Please try again." });
}
```

The `sendVerificationEmail` call at line 348 happens *after* the account creation at line 334. There is no rollback mechanism, no cleanup routine for orphaned accounts, and no way for the user to discover they can still verify via `/resend-otp`.

### Option A: Send Email First, Then Create User (Recommended)

Reverse the order of operations. If the email fails, no account exists yet, and the user can retry registration without conflict:

```typescript
// 1. Validate that email is not taken
const existingDbUser = await authRepository.findUserByEmail(email);
if (existingDbUser) {
  return res.status(409).json({ message: "An account with this email already exists." });
}

const passwordHash = hashPassword(password);
const otp = generateOtp();
const expiresAt = new Date(Date.now() + 10 * 60 * 1000);

// 2. Send verification email FIRST
const emailSent = await sendVerificationEmail(email, otp);
if (!emailSent) {
  return res.status(503).json({ message: "Failed to send verification email. Please try again." });
}

// 3. Only create account after email confirmed sent
const newUser = await authRepository.registerUserWithOtp(
  { fullName, email, medicalLicenseNumber: licenseNumber, passwordHash, emailVerified: false, role: "DOCTOR" },
  otp, expiresAt
);

// 4. Proceed with session creation
await regenerateSession(req);
req.session.pendingUser = { id: newUser.id, email };
await saveSession(req);
return res.status(201).json({ success: true, pendingEmail: email });
```

### Option B: Add Cleanup/Retry Logic

If email fails after account creation:
- Delete the just-created account (rollback)
- Return a user-friendly error with a retry suggestion

```typescript
const emailSent = await sendVerificationEmail(email, otp);
if (!emailSent) {
  // Rollback: delete the orphaned account
  await authRepository.deleteUserById(registeredUserId);
  return res.status(503).json({
    message: "Failed to send verification email. Please try again.",
    retry: true  // Client hint that re-registration will work
  });
}
```

### Option C: Improve Resend-OTP Discovery

If account creation succeeds but email fails, return a special response code that tells the client to redirect to the verification flow:

```typescript
if (!emailSent) {
  return res.status(202).json({
    success: true,
    pendingEmail: email,
    emailDeliveryFailed: true,
    message: "Account created but verification email could not be sent. Use /resend-otp to request a new code."
  });
}
```

ionfwsrijan:
## Summary
Fixes #2348 — The registration flow called `registerUserWithOtp()` before
`sendVerificationEmail()`. If the email service failed, an orphaned user
record with a pending OTP remained in the database.

## Changes
- `server/auth.ts`: Moved `sendVerificationEmail()` before
  `registerUserWithOtp()` so the user record is only created after the email
  is confirmed sent.

## Testing
- Verified that a failed email service does not create orphaned user records.
- Verified that successful registration still works correctly.

### DIFF
--- a/client/src/locales/en/translation.json
+++ b/client/src/locales/en/translation.json
@@ -262,28 +262,26 @@
       "descriptionNone": "Trend charts will appear here once assessment results are available for this account.",
       "actionLabel": "Refresh Trends"
     }
-  }
+  },
   "patientEducation": {
-  "title": "Patient Health Education",
-  "healthyDiet": "Maintain a balanced diet rich in vegetables, fruits, whole grains, and lean proteins.",
-  "physicalActivity": "Engage in at least 150 minutes of moderate physical activity every week.",
-  "weightManagement": "Maintain a healthy body weight to reduce diabetes-related complications.",
-  "bloodSugarMonitoring": "Regularly monitor your blood glucose levels and follow your healthcare provider's guidance.",
-  "medicationAdherence": "Take prescribed medications consistently and never stop treatment without medical advice.",
-  "regularCheckups": "Attend regular medical checkups to identify and prevent health complications early."
-},
-
-"followUpReminders": {
-  "title": "Medical Follow-up Reminders",
-  "appointment": "Schedule your next healthcare appointment on time for proper monitoring.",
-  "hba1cTest": "Repeat HbA1c testing according to your doctor's recommendation.",
-  "glucoseCheck": "Track your blood glucose regularly to monitor your diabetes risk.",
-  "lifestyleReview": "Review your diet, exercise routine, and lifestyle goals periodically."
-},
-
-"riskExplanation": {
-  "lowRisk": "Your current diabetes risk is low. Continue maintaining healthy lifestyle habits and routine screenings.",
-  "moderateRisk": "Your diabetes risk is moderate. Improving diet, increasing physical activity, and regular monitoring can help reduce future risk.",
-  "highRisk": "Your diabetes risk is high. Consult a healthcare professional for a detailed prevention and management plan."
-}
+    "title": "Patient Health Education",
+    "healthyDiet": "Maintain a balanced diet rich in vegetables, fruits, whole grains, and lean proteins.",
+    "physicalActivity": "Engage in at least 150 minutes of moderate physical activity every week.",
+    "weightManagement": "Maintain a healthy body weight to reduce diabetes-related complications.",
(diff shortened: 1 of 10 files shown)


ionfwsrijan:

PR #2377, opened 2026-07-13
fix(auth): preserve attemptCount across token replacements to prevent OTP brute-force via resend

### LINKED ISSUE #2351: [BUG] OTP Verification Exhaustion Attack via Token Replacement Bypass — Lockout Defeated by Resending Codes
## Description

The OTP verification flow implements a lockout mechanism that blocks further attempts after 3 failed verifications. However, this lockout is trivially bypassed by requesting a new OTP via the `/resend-otp` endpoint, which replaces the existing token (and its `attemptCount`) with a fresh token that resets `attemptCount` to 0.

An attacker who has captured a valid session can therefore brute-force a 6-digit OTP with effectively unlimited attempts:
1. Guess 3 OTP values → fail → locked out
2. Call `/resend-otp` → token replaced → `attemptCount` = 0
3. Guess 3 more OTP values → fail → locked out
4. Repeat step 2
5. Continue cycle until the correct OTP is found

For a 6-digit numeric OTP (1,000,000 possibilities), with the `/resend-otp` rate limiter allowing 3 requests per hour, the attacker gets 3 × 3 = 9 guesses per hour, which would require ~111,111 hours to exhaust the keyspace — not feasible. However, the lockout mechanism is **structurally defeated**: the attacker can never be permanently locked out, and the rate limiter is the only barrier.

The real risk is that the rate limiter at lines 82-88 allows 3 resend requests per hour. If an attacker:
- Has a pool of valid email addresses or sessions
- The rate limiter is per-IP but the OTP is per-user
- The attacker cycles between multiple users

This defeats the purpose of having an `attemptCount`-based lockout.

## Steps to Reproduce

1. Initiate login for `victim@example.com` and receive a pending session
2. Submit 3 incorrect OTP values → receive `429 Too Many Failed Attempts`
3. Call `POST /api/auth/resend-otp` with `{"email": "victim@example.com"}`
4. A new OTP is sent, and `attemptCount` resets to 0
5. Repeat steps 2-4 indefinitely — lockout is never permanent

## Expected Behavior

The lockout mechanism should be robust against token replacement:
- Failed attempt counts should persist across token replacements
- Or the account itself should have a cooldown period after N total failed attempts across all cycles
- Or resending OTP should increment a counter that has a hard limit

## Actual Behavior

Each `/resend-otp` call creates a fresh token with `attemptCount` reset to 0:

```typescript
// auth.ts:443
await authRepository.replaceVerificationToken(user.id, otp, expiresAt);
```

The underlying `replaceVerificationToken` creates a new `email_verification_tokens` row (or updates the existing one) with `attemptCount = 0`, completely erasing the failure history.

## Root Cause

Two architectural issues compound:

1. **Token-level locking vs. account-level locking**: Attempt counting is per-token, not per-user. When the token is replaced, the counter resets.

2. **Separation of lockout from rate limiting**: The `/resend-otp` endpoint's rate limiter (3 req/hour) is the only thing preventing infinite cycles, but this rate limiter is tied to the requester's IP/session, not to the target email account. An attacker with a botnet or IP rotation can bypass it entirely.

### Option A: Account-Level Failed Attempt Tracking (Recommended)

Add a `failed_otp_attempts` column to the `users` table and check it independently of the per-token count:

```typescript
// On OTP failure (auth.ts:516-525):
await tx
  .update(users)
  .set({ failedOtpAttempts: sql`COALESCE(failed_otp_attempts, 0) + 1` })
  .where(eq(users.id, user.id));

// Check before verification (auth.ts:503):
const userRecord = await authRepository.findUserById(user.id);
if ((userRecord.failedOtpAttempts ?? 0) >= 10) {
  return {
    success: false,
    status: 429,
    message: "Account temporarily locked due to too many failed attempts. Please try again in 30 minutes.",
  };
}
```

### Option B: Cumulative Attempt Tracking Across Token Replacements

Modify `replaceVerificationToken` to carry over the attempt count:

```typescript
async replaceVerificationToken(userId: string, otp: string, expiresAt: Date) {
  // Don't reset attemptCount — carry it forward
  const currentToken = await this.getCurrentToken(userId);
  const currentAttempts = currentToken?.attemptCount ?? 0;
  
  await db
    .insert(emailVerificationTokens)
    .values({
      userId,
      verificationCode: otp,
      expiresAt,
      attemptCount: currentAttempts, // Preserve existing count
    })
    .onConflictDoUpdate(...);
}
```

### Option C: Exponential Backoff on Resend

Implement an escalating cooldown that increases with each resend cycle:

```typescript
const resendAttempts = await getResendAttempts(email);
const cooldownMinutes = Math.min(2 ** resendAttempts, 120); // 1, 2, 4, 8, ... up to 120

if (Date.now() - lastResendAt < cooldownMinutes * 60 * 1000) {
  return res.status(429).json({
    message: `Please wait ${cooldownMinutes} minutes before requesting another code.`,
  });
}
```

### Option D: Combine All Three (Strongest)

- Account-level failed attempt counter with hard cap
- Exponential backoff on resend
- CAPTCHA requirement after 2 resend cycles

ionfwsrijan:
## Summary
Fixes #2351 — The `replaceVerificationToken` function in
`auth.repository.ts` invalidated the old token before creating the new one,
but did not carry over `attemptCount`. An attacker could exhaust OTP
attempts by triggering repeated token replacements.

## Changes
- `server/repositories/auth.repository.ts`: Modified `replaceVerificationToken`
  to read the `attemptCount` from the previous token before invalidation, and
  carry it forward to the new token.

## Testing
- Verified that `attemptCount` is preserved across token replacements.
- Verified that OTP exhaustion still triggers lockout after the correct number
  of attempts.

### DIFF
--- a/client/src/locales/en/translation.json
+++ b/client/src/locales/en/translation.json
@@ -262,28 +262,26 @@
       "descriptionNone": "Trend charts will appear here once assessment results are available for this account.",
       "actionLabel": "Refresh Trends"
     }
-  }
+  },
   "patientEducation": {
-  "title": "Patient Health Education",
-  "healthyDiet": "Maintain a balanced diet rich in vegetables, fruits, whole grains, and lean proteins.",
-  "physicalActivity": "Engage in at least 150 minutes of moderate physical activity every week.",
-  "weightManagement": "Maintain a healthy body weight to reduce diabetes-related complications.",
-  "bloodSugarMonitoring": "Regularly monitor your blood glucose levels and follow your healthcare provider's guidance.",
-  "medicationAdherence": "Take prescribed medications consistently and never stop treatment without medical advice.",
-  "regularCheckups": "Attend regular medical checkups to identify and prevent health complications early."
-},
-
-"followUpReminders": {
-  "title": "Medical Follow-up Reminders",
-  "appointment": "Schedule your next healthcare appointment on time for proper monitoring.",
-  "hba1cTest": "Repeat HbA1c testing according to your doctor's recommendation.",
-  "glucoseCheck": "Track your blood glucose regularly to monitor your diabetes risk.",
-  "lifestyleReview": "Review your diet, exercise routine, and lifestyle goals periodically."
-},
-
-"riskExplanation": {
-  "lowRisk": "Your current diabetes risk is low. Continue maintaining healthy lifestyle habits and routine screenings.",
-  "moderateRisk": "Your diabetes risk is moderate. Improving diet, increasing physical activity, and regular monitoring can help reduce future risk.",
-  "highRisk": "Your diabetes risk is high. Consult a healthcare professional for a detailed prevention and management plan."
-}
+    "title": "Patient Health Education",
+    "healthyDiet": "Maintain a balanced diet rich in vegetables, fruits, whole grains, and lean proteins.",
+    "physicalActivity": "Engage in at least 150 minutes of moderate physical activity every week.",
+    "weightManagement": "Maintain a healthy body weight to reduce diabetes-related complications.",
+    "bloodSugarMonitoring": "Regularly monitor your blood glucose levels and follow your healthcare provider's guidance.",
+    "medicationAdherence": "Take prescribed medications consistently and never stop treatment without medical advice.",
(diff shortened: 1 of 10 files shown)


ionfwsrijan:

